PR #49, opened 2014-05-17
Generate shorter index names so as not to fail on PostgreSQL or MySQL (or others?) with long table names

### LINKED ISSUE #48: Transition table migration fails with long table name on MySQL
```
-- add_index(:internet_payment_transitions, [:sort_key, :internet_payment_id], {:unique=>true})
rake aborted!
An error has occurred, all later migrations canceled:

Index name 'index_internet_payment_transitions_on_sort_key_and_internet_payment_id' on table 'internet_payment_transitions' is too long; the limit is 64 characters
/home/liam/.rvm/gems/ruby-2.0.0-p451/gems/activerecord-3.2.12/lib/active_record/connection_adapters/abstract/schema_statements.rb:573:in `a
dd_index_options'
...
```

Clearly the issue is that Rails generates a default index name that's too long, so the solution would be to explicitly pass in a shorter index name to the migration template. 

I'm working on a fix for this now.

hut8:
I don't see any tests for the generators (I could be blind), so... hope this works!

Should fix #48

### DIFF
--- a/lib/generators/statesman/active_record_transition_generator.rb
+++ b/lib/generators/statesman/active_record_transition_generator.rb
@@ -33,6 +33,10 @@ module Statesman
       klass.underscore.pluralize
     end
 
+    def index_name(index_id)
+      "index_#{table_name}_#{index_id}"
+    end
+
     def parent_id
       parent.underscore + "_id"
     end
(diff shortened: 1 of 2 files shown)


asmaloney:
Just ran into this problem.  @hut8 Any chance of finishing up and merging this?  Thanks!

greysteil:
I hate to see a good PR go stale - have rebased this and added a spec here: https://github.com/gocardless/statesman/pull/127

@hut8 - your commit is still included, so you'll still get all credit.

Issue #59, opened 2014-06-07
Optimize MySQL indexes?


mkdynamic:
Currently, there are 2 indexes added [by default in the migration for the ActiveRecord transition model](https://github.com/gocardless/statesman/blob/master/lib/generators/statesman/templates/create_migration.rb.erb#L11). Since all queries will always include the [foreign key for the model](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L56), I believe it would be more optimal to use 1 index.

```ruby
# Current indexes:
add_index :model, :model_id
add_index :model, [:sort_key, :model_id], unique: true

# Suggested index:
add_index :model, [:model_id, :sort_key], unique: true
```

Assuming that the query to obtain the transition history always includes the foreign key (`model_id`) and is sorted by `sort_key`, this seems optimal for 2 reasons:

1. Reduces from 2 indexes to 1, which will reduce insert time
2. Because the `sort_key` comes last, I believe MySQL can use the index to sort the results, rather than having to perform a file sort

Some quick EXPLAIN analysis seems to confirm 2.

Does this seem sane? Let me know I misunderstood/overlooked anything.

greysteil:
Closing now https://github.com/gocardless/statesman/pull/92 is merged.

PR #92, opened 2014-10-12
Optimize MySQL indexes for transition model

mkdynamic:
See https://github.com/gocardless/statesman/issues/59 for more details.

### DIFF
--- a/lib/generators/statesman/templates/create_migration.rb.erb
+++ b/lib/generators/statesman/templates/create_migration.rb.erb
@@ -8,7 +8,6 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.timestamps
     end
 
-    add_index :<%= table_name %>, :<%= parent_id %>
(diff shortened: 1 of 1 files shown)


greysteil:
Looks good to me. @hmarr / @alan / @Sinjo - can one of you sign off?

http://dev.mysql.com/doc/refman/5.0/en/multiple-column-indexes.html
http://www.postgresql.org/docs/9.0/static/indexes-multicolumn.html

Sinjo:
This makes sense. I can't imagine a case where you'd use the `sort_key` in a way that isn't scoped to a specific parent, so there's no advantage to `[:sort_key, :parent_id]` over `[:parent_id, :sort_key]`.

When you're only looking up by the `parent_id`, Postgres will happily use the first column of the index (and I believe the same holds in other databases).

greysteil:
@mkdynamic - thanks for this, and apologies for the delay getting it merged!

mkdynamic:
No worries. Thanks!

Issue #93, opened 2014-10-15
DEPRECATION WARNING when using in_state and not_in_state


netikular:
I'm using rails 4.0.8 and receiving: 

DEPRECATION WARNING: It looks like you are eager loading table(s) (one of: requests, transition1, transition2, floor) that are referenced in a string SQL snippet.

I've resolved the problem on a copy of my own, I'll submit a pull request to discuss if this is the way you would like to fix it.

greysteil:
Closing since https://github.com/rails/rails/commit/d1e7cd14c29f1ef45f2f36e79cd99eb580036991 removes this deprecation.

greysteil:
@netikular - thanks all the same for this, and sorry it's taken us this long to get to the bottom of it!

PR #94, opened 2014-10-15
Resolving deprecation warnings when using implicit joins

netikular:
Hi, I was getting a warning on recent versions of rails. 

DEPRECATION WARNING: It looks like you are eager loading table(s) (one of: requests, transition1, transition2, floor) that are referenced in a string SQL snippet.

I looked into the code and noticed that the in_state and the not_in_state scopes do not reference the transition table.  I've added the reference and the include to query and that resolved the deprecation warning I was getting.  I also submitted issue #93 to track a conversation about this.

Since this is not coming up in the specs when run on just the project itself, I am not sure how to add a spec for this.

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -9,19 +9,23 @@ module Statesman
         def in_state(*states)
           states = states.map(&:to_s)
 
-          joins(transition1_join)
+          includes(transition_name)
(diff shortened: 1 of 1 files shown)


greysteil:
Hey @netikular, looks like this deprecation has since been aborted - take a look at https://github.com/rails/rails/commit/d1e7cd14c29f1ef45f2f36e79cd99eb580036991

PR #111, opened 2014-12-18
Support for and testing against pre-4.1 Rails

isaacseymour:
`ActiveRecord.gem_version` was introduced in 4.1, and it seems to be the only thing stopping us from supporting older Rails versions. Added 3.2 and 4.0 to the build matrix and altered usage of `ActiveRecord.gem_version` to work for pre-4.1. @greysteil could you review? I don't remember there being a specific decision to drop old-Rails support, if there was we should document it instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,7 @@ script:
  - bundle exec rubocop
  - bundle exec rake
 env:
-  - "RAILS_VERSION=4.1.7"
-  - "RAILS_VERSION=4.2.0.rc2"
+  - "RAILS_VERSION=3.2.21"
+  - "RAILS_VERSION=4.0.12"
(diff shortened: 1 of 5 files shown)


isaacseymour:
The problem with rails 4.0/ruby 1.9.3 had something to do with not specifying a default for metadata in the specs. I'll squash and rebase against master (which has the rubocop fixes in).

isaacseymour:
(would be good to test against Ruby 2.2 for Rails 4.1+, but I can't work out how to specify a build matrix like that)

isaacseymour:
@jackfranklin can you review now the specs are :grinning: ?

jackfranklin:
:+1:

PR #112, opened 2014-12-19
Return transition object on successful transition

div:
At first I wanted to ask if there're any reasons to return 'true' instead of the transition object, but it was so easy to change, so I went with a PR. 

It's easy to use history.last, but can lead to race conditions, so i think it's a good idea to return exactly what has been created by the transition. 

I'll be happy to change the readme if you decide this change is fine. Great gem btw! :heart: Thanks!


petehamilton:
Hey Igor, we’ve just been considering potential breaking changes in statesman v2.0 and this one came up.

After thinking it through, we'd actually prefer to keep the explicit `[true|false]` return values, rather than `[object|nil]`, given that you can already retrieve the last transition easily via `machine.last_transition`, which is a cached and therefore a cheap additional call.

In your example, I think this would look like:

```ruby
class Issues::ClosingsController < ApplicationController
  def create
    if @issue.trigger!(:close, closed_by_id: current_user.id)
      render json: @issue.last_transition, status: :created
    else
      render json: { error: :unclosed }, status: :unprocessable_entity
    end
  end
end
```

What do you think?

petehamilton:
Hi @div any thoughts on the above?

div:
Hey Pete! Sorry for delay, haven't seen the first notification! 

I know that i can use last_transition sure, if you think [true, false] is the preferred way, than go for it!
And thx again for the gem!

hale:
@petehamilton  Forgive me for digging up an old issue, but I don't think using `last_transition` is a good idea because of the potential for race conditions that @div mentioned.

At least in the ActiveRecord storage adapter, [calling `transition_to` wipes the cache](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L41), causing an additional query to run when asking for the most recent transition.

The space between the two transactions (INSERTing the transition and SELECTing the most recent) is where a race condition can occur.

danwakefield:
@hale Unfortunately, I don't think we'll be making a breaking change in behaviour  anytime soon.

I would recommend patching the machine class in an initializer with
```
module Statesman
  module Machine
    def transition_to!(new_state, metadata = {})
      initial_state = current_state
      new_state = new_state.to_s

      validate_transition(from: initial_state,
                          to: new_state,
                          metadata: metadata)

      @storage_adapter.create(initial_state, new_state, metadata)
      
      # We have removed the return of `true` so we have access to the
      # transition created by `transition_to!`
      # true
    end
  end
end
```

This would then mean you get the created transition for `transition_to!` and `transition_to!` if they are successful. 

I can't be certain that the transition will be totally up to date as we don't reload it after saving it so I'd add testing around this before rolling it out.

PR #114, opened 2015-01-06
Look at the TLDR example using memoization - is this valid?

PragTob:
It seemed odd to me that this value is not memoized but instead
the object is recreated on every access. Is this intentional
because memoization (for some reason) does not work here or does
memoization work as well?

(Decided to PR as this is easier to discuss and maybe easier to
 fix then)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ class Order < ActiveRecord::Base
   has_many :order_transitions
 
   def state_machine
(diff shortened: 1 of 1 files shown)


isaacseymour:
Good spot! I'm actually pretty sure there could well be unexpected behaviour without memoization.

Issue #119, opened 2015-01-23
Deprecation warning in Rails 4.2


moomerman:
Just spotted this warning during transitions table migration

```
> -- create_table(:show_transitions)

! DEPRECATION WARNING: `#timestamp` was called without specifying an option for `null`. 
In Rails 5, this behavior will change to `null: false`. 
You should manually specify `null: true` to prevent the behavior of your existing migrations from changing.
```

isaacseymour:
Interesting, I'd assumed that was the default behaviour in Rails 4. Thanks for pointing this out, #120 should fix it.

isaacseymour:
#120 is merged, so this will be fixed in the next release. Thanks for pointing this out!

PR #120, opened 2015-01-23
Specify null: false on timestamps for Rails 5 readiness

isaacseymour:
For #119 

@jackfranklin I think timestamps should probably always be `NOT NULL`, otherwise they're a bit pointless.


jackfranklin:
+1

PR #123, opened 2015-02-11
add metadata parameter to allowed_transitions method

pilu:
This might be useful to check available transitions with guards, like checking valid transitions for the current user role.

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -192,9 +192,9 @@ module Statesman
       last_action ? last_action.to_state : self.class.initial_state
     end
 
-    def allowed_transitions
+    def allowed_transitions(metadata = {})
       successors_for(current_state).select do |state|
-        can_transition_to?(state)
+        can_transition_to?(state, metadata)
       end
     end
 
--- a/spec/statesman/adapters/shared_examples.rb
+++ b/spec/statesman/adapters/shared_examples.rb
@@ -67,8 +67,7 @@ shared_examples_for "an adapter" do |adapter_class, transition_class|
     context "with after callbacks" do
       it "is called after the state transition" do
         expect(observer).to receive(:execute)
-          .with(:after, anything, anything, anything) {
-            |_phase, _from_state, _to_state, transition|
+          .with(:after, anything, anything, anything) { |_, _, _, transition|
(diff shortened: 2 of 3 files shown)


isaacseymour:
This seems pretty sensible - @barisbalic @Sinjo @jackfranklin what do you think? Would need to add some tests to make sure it works before merging.

jackfranklin:
No objections to the change from me.

pilu:
@isaacseymour I added some tests, what do you think?

Sinjo:
Good addition to the API. :+1: from me.

jackfranklin:
@pilu looks really good to me. Will let @isaacseymour make the final call but big +1. Thanks :D

pilu:
@jackfranklin cool, thank you guys for this gem :metal:

isaacseymour:
I'm :+1: - will merge if @barisbalic is :+1: on the API change

barisbalic:
@pilu thanks very much for raising this PR, it prompted some serious internal discussion about an aspect of Statesman we had taken for granted.  In the current context everything you've suggested makes sense and the code is good, but we're going to have to decline this PR.

It's become clear that we should have been more careful about passing metadata around; it makes a lot of sense to be able to store context about transitions, but it also feels like a really bad idea to alter behaviour based on this somewhat "coincidental", loosely attached data.

If there's some piece of information that could change the outcome of a transition, it should be explicitly attached (as a field) to the transition itself, or more likely the class/model that houses the state machine.  In the context of your user role example, from within a guard clause, if you want to check a user role, it makes more sense to consult the user/permissions structure than it does to pull something out of metadata.  We're making use of Service objects in these types of situations, so it would look something like...

```
guard_transition(to: :deleted) do |item|
  user_access_control.can_delete_item?(current_user, item)
end
```

As a result of this debate, we're going to remove the argument from the `can_transition_to?` in the next version of Statesman, which should be coming in the next month or two.

Thanks again for your efforts on this, they are greatly appreciated and have pushed us to make Statesman a better state machine.  If you have any further feelings or thoughts on the matter please continue to comment, but for now I'll close the thread.

pilu:
@barisbalic cool thank you very much for the explanation, it makes sense!

PR #125, opened 2015-02-14
Allow ActiveRecordQueries to accept array as an argument.

robwilliams:
Hopefully this fits within your guidelines.. 

I am using these query methods with has_scope which forwards an array as an argument to the related query method.

Allows the following usage:

* `.in_state([:succeeded, :failed])`
* `.not_in_state([:succeeded, :failed])`

This matches the behaviour of ActiveRecord::Base#find where `Person.find(1, 2, 6)` or
`Person.find([1, 2, 6])` will return the same result.

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -7,7 +7,7 @@ module Statesman
 
       module ClassMethods
         def in_state(*states)
-          states = states.map(&:to_s)
+          states = states.flatten.map(&:to_s)
 
           joins(transition1_join)
             .joins(transition2_join)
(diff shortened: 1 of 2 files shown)


isaacseymour:
I'm :+1:

@barisbalic @Sinjo @mrappleton what do you guys think?

appleton:
Looks good to me!

Looks like CI is failing on a rubocop error which is unrelated to the change
business-man-shrug-quesiton-mark-279604

isaacseymour:
Yep, this keeps happening - I think maybe our rubocop version isn't locked
down? I'll fix up master later

Sinjo:
I'm :+1: on this too!

Issue #130, opened 2015-03-03
Namespaced Models


DanielWright:
Hi there,

I'm having some difficulty implementing Statesman with namespaced ActiveRecord models. The root cause, I've discovered, is the Stateman ActiveRecord adapter [relies](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L71-L73) on the transition model's `table_name` exactly matching its has_many association to the parent model.

In the following example, this falls apart:

```ruby
module MyNamespace
  class Foo < AR::Base
    def self.table_name_prefix
      "my_namespace_"
    end

    has_many :foo_transitions, class_name: "MyNamespace::FooTransition"

   def state_machine
     @_machine ||= Statesman::Machine.new(self, transition_class: MyNamespace::FooTransition)
   end
  end

  class FooTransition < AR::Base
    include Statesman::Adapters::ActiveRecordTransition

    def self.table_name_prefix
      "my_namespace_"
    end

    belongs_to :foo, class_name: 'MyNamespace::Foo',
      inverse_of: :foo_transitions
  end
end

$ bundle exec rails console
2.2.0 > MyNamespace::Foo.last.state_machine.history
NoMethodError: undefined method `my_namespace_foo_transitions' for #<MyNamespace::Foo id: 7>
```

I considered proposing a pull request that reflects on the parent model's associations and tries to infer the transitions association via tortured comparison with `@transition_class.model_name`. However, I realized even that inference would break down if the transition-class's table-name doesn't match the association for whatever reason.

I propose instead adding another configuration option to `Statesman::Machine`'s initializer, something like `association_name`, which can be passed down to the adapter and sent to the parent model to retrieve the transitions collection. I'm working on a PR, but I'd appreciate any feedback on the idea before I go too far down that road.

Thanks!

isaacseymour:
Hi @DanielWright, the second approach sounds sensible. I'd be in favour of keeping the current behaviour as default (it's correct 99% of the time), but having the `association_name` available as an override. Look forward to the PR!

greysteil:
Closing now #131 is merged and included in a release. Thanks again @DanielWright !

PR #131, opened 2015-03-03
Support for Unconventional Transition Associations

DanielWright:
This pull request adds support for an `association_name` option when initializing a state-machine. This supports transition model associations where the transition model's table-name doesn't match the association--for example, when models are namespaced.

I left in some rough specs that motivated the change in implementation. They don't provide much value, and I think they can reasonably be removed, but let me know what your preference is.

See #130 for motivation and other details.

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -66,23 +66,31 @@ module Statesman
           transition_class.table_name.to_sym
         end
 
+        def transition_reflection
+          reflect_on_association(transition_name)
+        end
+
         def model_foreign_key
-          reflect_on_association(transition_name).foreign_key
+          transition_reflection.foreign_key
+        end
+
+        def model_table
+          transition_reflection.table_name
         end
 
         def transition1_join
-          "LEFT OUTER JOIN #{transition_name} transition1
+          "LEFT OUTER JOIN #{model_table} transition1
              ON transition1.#{model_foreign_key} = #{table_name}.id"
         end
 
         def transition2_join
-          "LEFT OUTER JOIN #{transition_name} transition2
+          "LEFT OUTER JOIN #{model_table} transition2
              ON transition2.#{model_foreign_key} = #{table_name}.id
              AND transition2.sort_key > transition1.sort_key"
         end
 
         def most_recent_transition_join
-          "LEFT OUTER JOIN #{transition_name} AS last_transition
+          "LEFT OUTER JOIN #{model_table} AS last_transition
              ON #{table_name}.id = last_transition.#{model_foreign_key}
              AND last_transition.most_recent = #{db_true}"
         end
--- a/spec/statesman/adapters/active_record_spec.rb
+++ b/spec/statesman/adapters/active_record_spec.rb
@@ -204,4 +204,25 @@ describe Statesman::Adapters::ActiveRecord, active_record: true do
       end
     end
   end
+
+  context "with a namespaced model" do
+    before do
+      silence_stream(STDOUT) do
+        CreateNamespacedARModelMigration.migrate(:up)
+        CreateNamespacedARModelTransitionMigration.migrate(:up)
+      end
+    end
+
+    before do
+      MyNamespace::MyActiveRecordModelTransition.serialize(:metadata, JSON)
+    end
+    let(:observer) { double(Statesman::Machine, execute: nil) }
+    let(:model) do
+      MyNamespace::MyActiveRecordModel.create(current_state: :pending)
+    end
+
+    it_behaves_like "an adapter",
+                    described_class, MyNamespace::MyActiveRecordModelTransition,
+                    association_name: :my_active_record_model_transitions
+  end
 end
--- a/spec/statesman/adapters/shared_examples.rb
+++ b/spec/statesman/adapters/shared_examples.rb
@@ -11,9 +11,16 @@ require "spec_helper"
 #   history:          Returns the full transition history
 #   last:             Returns the latest transition history item
 #
-shared_examples_for "an adapter" do |adapter_class, transition_class|
+# rubocop:disable Metrics/LineLength
+# NOTE This line cannot reasonably be shortened.
+shared_examples_for "an adapter" do |adapter_class, transition_class, options = {}|
+  # rubocop:enable Metrics/LineLength
+
   let(:observer) { double(Statesman::Machine, execute: nil) }
-  let(:adapter) { adapter_class.new(transition_class, model, observer) }
+  let(:adapter) do
+    adapter_class.new(transition_class,
+                      model, observer, options)
+  end
 
   describe "#initialize" do
     subject { adapter }
--- a/spec/support/active_record.rb
+++ b/spec/support/active_record.rb
@@ -37,7 +37,7 @@ class CreateMyActiveRecordModelMigration < ActiveRecord::Migration
   def change
     create_table :my_active_record_models do |t|
       t.string :current_state
-      t.timestamps(null: false)
+      t.timestamps null: false
     end
   end
 end
@@ -59,7 +59,7 @@ class CreateMyActiveRecordModelTransitionMigration < ActiveRecord::Migration
         t.text :metadata, default: '{}'
       end
 
-      t.timestamps(null: false)
+      t.timestamps null: false
     end
 
     add_index :my_active_record_model_transitions,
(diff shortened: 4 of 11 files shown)


.ruby-version
```diff
@@ -0,0 +1 @@
+2.1.5
```
isaacseymour: Can you remove this from the PR?

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -148,4 +148,25 @@
...
+
+      silence_stream(STDOUT) do
+        CreateMyNamespaceMyActiveRecordModelTransitionMigration.migrate(:up)
+      end
+    end
+
+    before { MyNamespace::MyActiveRecordModelTransition.serialize(:metadata, JSON) }
+    let(:observer) do
+      result = double(Statesman::Machine)
+      allow(result).to receive(:execute)
```
isaacseymour: As a style point I'd prefer
```ruby
let(:observer) { double(Statesman::Machine, execute: nil) }
```

.ruby-version
DanielWright: Done.

spec/statesman/adapters/shared_examples.rb
```diff
@@ -11,13 +11,13 @@
 #   history:          Returns the full transition history
 #   last:             Returns the latest transition history item
 #
-shared_examples_for "an adapter" do |adapter_class, transition_class|
+shared_examples_for "an adapter" do |adapter_class, transition_class, association_name = nil|
```
isaacseymour: Can this become `|adapter_class, transition_class, options = {}|` instead so it's the same as the initializer? This line could get very long if more options get added otherwise!

spec/support/active_record.rb
```diff
@@ -55,3 +55,58 @@ def change
...
+  class MyActiveRecordModel < ActiveRecord::Base
+    has_many :my_active_record_model_transitions,
+      class_name: "MyNamespace::MyActiveRecordModelTransition"
+
+    def self.table_name_prefix
+      "my_namespace_"
+    end
+
+    def state_machine
+      @_state_machine ||= MyStateMachine.new(self,
```
isaacseymour: `@state_machine` instead of `@_state_machine` please!

spec/support/active_record.rb
```diff
@@ -55,3 +55,58 @@ def change
...
+end
+
+class CreateMyNamespaceMyActiveRecordModelTransitionMigration < ActiveRecord::Migration
+  def change
+    create_table :my_namespace_my_active_record_model_transitions do |t|
+      t.string  :to_state
+      t.integer :my_active_record_model_id
+      t.integer :sort_key
+      t.text    :metadata, default: '{}'
+      t.timestamps(null: false)
```
isaacseymour: Drop the `()` for consistency with the rest of the migration

isaacseymour:
I think maybe this needs some more specs, although I'm not that strong on ActiveRecord stuff. 

@jackfranklin @Sinjo @barisbalic what do you guys think?

spec/statesman/adapters/active_record_spec.rb
Sinjo: `instance_double` would be even nicer (verifies stubbed methods exist).

jackfranklin:
Specs look OK to me and change seems like a sensible one :+1:

DanielWright:
Forgot to check rubocop—whoops! I'll get on that shortly.

However, I've discovered another area that relies on the table-name and association-name matching exactly:  [`Statesman::Adapters::ActiveRecordQueries#transition_name`](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record_queries.rb#L39-L41). I'm working on a refactoring that will address this and I'll attach it to this PR, if that's okay.

DanielWright:
What's your preferred squashing method? Would you like me to preserve the commit messages?

DanielWright:
4cac51b and f4b441d are my attempts at resolving the hidden dependency on matching association- and table-names. Let me know what you think. I don't know what your attitude(s) are re: deprecations.

isaacseymour:
What's the problem with just overriding the `transition_name` method if you're namespacing your transition model? I'm not a fan of any change which introduces deprecation warnings without a really really good reason.

barisbalic:
Big fan of your avatar.  Have to agree with Isaac, seems like a bad idea to base functionality on features that are being removed.

DanielWright:
Wow, yup, that's a much better solution. I lost sight of the forest for the trees on that.

barisbalic:
That's cool, it happens to all of us.  Thanks for taking the time to flesh out your proposal so thoroughly, I'll close this issue for now, feel free to re-open it if something else pops up.

barisbalic:
Misread, un-closificating.

DanielWright:
Hah, yes, sorry if I was vague—not much sleep to be had last night, and my brain is not its sharpest.

DanielWright:
Given the adjustments I committed last week, does this PR look more reasonable?

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -9,11 +9,7 @@
   end
 
   before { MyActiveRecordModelTransition.serialize(:metadata, JSON) }
-  let(:observer) do
-    result = double(Statesman::Machine)
-    allow(result).to receive(:execute)
-    result
-  end
+  let(:observer) { double(Statesman::Machine, execute: nil) }
```
isaacseymour: Very :+1: on this!

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -148,4 +144,28 @@
       end
     end
   end
+
+  context "with a namespaced model" do
+    before do
+      silence_stream(STDOUT) do
+        CreateNamespacedARModelMigration.migrate(:up)
+      end
```
isaacseymour: Squash these into a single `silence_stream` block?

spec/statesman/adapters/shared_examples.rb
```diff
@@ -11,13 +11,20 @@
...
 #
-shared_examples_for "an adapter" do |adapter_class, transition_class|
+# rubocop:disable Metrics/LineLength
+# NOTE This line cannot reasonably be shortened.
+shared_examples_for "an adapter" do |adapter_class, transition_class, options = {}|
+  # rubocop:enable Metrics/LineLength
+
   let(:observer) do
     result = double(Statesman::Machine)
     allow(result).to receive(:execute)
```
isaacseymour: Could do the same `double(Statesman::Machine, execite: nil)` trick here as above if you fancy changing it :P

isaacseymour:
I'm :+1: once this is squashed and rebased. Sorry for the delay @DanielWright

DanielWright:
Squashed and rebased, as requested! Thanks for the attention, @isaacseymour.

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -66,17 +66,25 @@ def transition_name
...
 
         def transition1_join
-          "LEFT OUTER JOIN #{transition_name} transition1
+          "LEFT OUTER JOIN #{model_table} transition1
              ON transition1.#{model_foreign_key} = #{table_name}.id"
         end
 
         def transition2_join
-          "LEFT OUTER JOIN #{transition_name} transition2
+          "LEFT OUTER JOIN #{model_table} transition2
```
isaacseymour: I think you need to apply this change to the new `most_recent_transition_join` method too to make this work with the `most_recent` column added in 1.2.0

isaacseymour:
I'm :+1: after the comment, would like a :+1:  from @jackfranklin / @barisbalic / @alan before merging.

barisbalic:
Pending the comment I'm :+1:, thanks for your hard work @DanielWright

alan:
Looks good to me! Would be great to get the Readme updated showing how to use this.

lib/statesman/adapters/active_record_queries.rb
DanielWright: Ah, good catch.

DanielWright:
Let me know if this latest update looks good. I addressed the comment, plus updated the README.

isaacseymour:
Looks good to me. Second :+1: on the README changes would be good @jackfranklin @barisbalic @alan

alan:
great! thanks :+1:

jackfranklin:
Thanks for the contrib @DanielWright ! :+1: from me

isaacseymour:
Actually, looks like this doesn't work with MySQL - https://travis-ci.org/gocardless/statesman/builds/55673316 (I don't know why Travis isn't webhooking the results of builds back to GitHub...). Hopefully that's not too hard to fix?

DanielWright:
Yeah, it was a question of not assigning a default value to the `metadata` column [in the namespaced migration](https://github.com/DanielWright/statesman/commit/eb6729f15a1936d758abe6ecb53bf94a26ecf259#diff-8e8ab0435d5fe510d68fdcc013ac08a9R133). CI build should pass now!

isaacseymour:
Woop, will merge when travis is happy

isaacseymour:
Pushed as v1.2.2. Thanks again for all the work and patience, @DanielWright

DanielWright:
My pleasure. Thanks for the feedback and for a great gem.

PR #132, opened 2015-03-09
Add most_recent column to transitions

greysteil:
For discussion. We're using something similar to this at GoCardless to speed up queries on our larger tables.

This PR adds a binary `most_recent` column to the `_transitions` table, and automatically updates that column each time an object is transitioned. This allows `in_state` and `not_in_state` queries to only consider rows where the `most_recent` flag is true, and removes the need to join the transitions table on itself (the current statesman method for identifying the most recent transition).

This PR is fully backwards compatible with the existing statesman query implementation - if a `most_recent` column isn't present it will fall back to the old implementation. To make it really awesome, however, we should include details of how to back-fill old transitions.

### DIFF
--- a/lib/generators/statesman/templates/create_migration.rb.erb
+++ b/lib/generators/statesman/templates/create_migration.rb.erb
@@ -5,9 +5,11 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.text :metadata<%= ", default: \"{}\"" unless mysql? %>
       t.integer :sort_key, null: false
       t.integer :<%= parent_id %>, null: false
+      t.boolean :most_recent, null: false
       t.timestamps null: false
     end
 
-    add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort  %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :most_recent], unique: true, where: "most_recent", name: "<%= index_name :parent_most_recent %>"
   end
 end
--- a/lib/generators/statesman/templates/update_migration.rb.erb
+++ b/lib/generators/statesman/templates/update_migration.rb.erb
@@ -8,5 +8,6 @@ class AddStatesmanTo<%= migration_class_name %> < ActiveRecord::Migration
     add_column :<%= table_name %>, :updated_at, :datetime, null: false
 
     add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort  %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :most_recent], unique: true, where: "most_recent", name: "<%= index_name :parent_most_recent %>"
   end
 end
--- a/lib/generators/statesman/templates/update_migration.rb.erb
+++ b/lib/generators/statesman/templates/update_migration.rb.erb
@@ -4,6 +4,7 @@ class AddStatesmanTo<%= migration_class_name %> < ActiveRecord::Migration
     add_column :<%= table_name %>, :metadata, :text<%= ", default: \"{}\"" unless mysql? %>
     add_column :<%= table_name %>, :sort_key, :integer, null: false
     add_column :<%= table_name %>, :<%= parent_id %>, :integer, null: false
+    add_column :<%= table_name %>, :most_recent, null: false
     add_column :<%= table_name %>, :created_at, :datetime, null: false
     add_column :<%= table_name %>, :updated_at, :datetime, null: false
 
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -27,8 +27,8 @@ module Statesman
         to = to.to_s
         create_transition(from, to, metadata)
       rescue ::ActiveRecord::RecordNotUnique => e
-        if e.message.include?('sort_key') &&
-           e.message.include?(@transition_class.table_name)
+        if e.message.include?(@transition_class.table_name) &&
+           e.message.include?('sort_key') || e.message.include?('most_recent')
           raise TransitionConflictError, e.message
         else raise
         end
@@ -53,12 +53,17 @@ module Statesman
       private
 
       def create_transition(from, to, metadata)
-        transition = transitions_for_parent.build(to_state: to,
-                                                  sort_key: next_sort_key,
-                                                  metadata: metadata)
+        transition_attributes = { to_state: to,
+                                  sort_key: next_sort_key,
+                                  metadata: metadata }
+
+        transition_attributes.merge!(most_recent: true) if most_recent_column?
+
+        transition = transitions_for_parent.build(transition_attributes)
 
         ::ActiveRecord::Base.transaction do
           @observer.execute(:before, from, to, transition)
+          unset_old_most_recent
           transition.save!
           @last_transition = transition
           @observer.execute(:after, from, to, transition)
@@ -70,6 +75,15 @@ module Statesman
 
       def transitions_for_parent
         @parent_model.send(@transition_class.table_name)
+      end
+
+      def unset_old_most_recent
+        return unless most_recent_column?
+        transitions_for_parent.update_all(most_recent: false)
+      end
+
+      def most_recent_column?
+        transition_class.columns_hash.include?("most_recent")
       end
 
       def next_sort_key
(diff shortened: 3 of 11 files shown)


hmarr:
A few additional points worth mentioning:

- This problem could also be solved by serialising the current state onto the parent. However, we'd rather keep Statesman-specific fields on the transitions table and leave the parent model be.
- We also considered using `valid_{to,from}` fields rather than `most_recent`, but this makes the transition logic more complex. The advantage of using `valid_{from,to}` is that historic queries become a bit easier.
- We keep the size of transition table indexes manageable by using partial indexes (`CREATE INDEX ... WHERE most_recent = 't'`). We also maintain a unique partial index over `parent_id, most_recent` to guard against bad data. 
- The auto-updating `most_recent` on transition can also be done with a trigger (we currently do this internally) - see below.

```sql
CREATE FUNCTION new_foo_transition_before() RETURNS trigger
LANGUAGE plpgsql
AS $$
BEGIN
  UPDATE foo_transitions SET most_recent = 'f' WHERE foo_id = NEW.foo_id AND most_recent = 't';
  RETURN NEW;
END;
$$;
```

greysteil:
@hmarr @petehamilton @myitcv - having written this up for statesman, I'm not madd keen on doing the auto-updating in a create callback (database or app code level). Is there a reason we didn't put it into a callback on the state machine itself? That way any back-fill tasks don't have to worry about accidentally triggering the callback.

hmarr:
Can't remember exactly why (I was only briefly in the discussion), but the main advantage to using a trigger is you can transition reliably from outside of statesman (e.g. bulk-transitioning from SQL). Also probably reduces lock contention on concurrent updates. Downside is back-population requires disabling triggers (I think this can be done by setting the session's replication role). @Sinjo, thoughts?

hmarr:
Also, avoiding triggers may make life easier for non-SQL backends.

myitcv:
@hmarr 100% agree with all the advantages you list. However, the trigger is only fired on `INSERT` hence the back population doesn't require it (the trigger) to be disabled; the state table is, in effect, append-only:

```
CREATE TRIGGER foo_transitions_before BEFORE INSERT ON foo_transitions FOR EACH ROW EXECUTE PROCEDURE new_foo_transition_before();
```

hmarr:
@myitcv back-filling data for a new transition type would require inserts, wouldn't it? E.g. when introducing "submitted" between "created" and "paid", you'd need to insert the new "submitted" transition for the existing data.

myitcv:
@hmarr you're absolutely correct, my bad. I was recalling how we migrated to this model (which needed the backfilling of `most_recent = 'f'`, which results only in an `UPDATE`). The case of a introducing a new transition definitely needs to be considered.

greysteil:
This is now ready for a proper review. I've updated this to do the update of `most_recent` in the state transition rather than on a callback and it feels a lot cleaner. Tempted to update the migration so that there's no default, too.

@petehamilton and @hmarr - can you do the honours?

lib/generators/statesman/templates/create_migration.rb.erb
```diff
@@ -5,9 +5,11 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.text :metadata<%= ", default: \"{}\"" unless mysql? %>
       t.integer :sort_key, null: false
       t.integer :<%= parent_id %>, null: false
+      t.string :most_recent, default: 't', limit: 1, null: false
```
hmarr: `boolean` rather than `string`

lib/statesman/adapters/active_record.rb
```diff
@@ -59,6 +59,7 @@ def create_transition(from, to, metadata)
 
         ::ActiveRecord::Base.transaction do
           @observer.execute(:before, from, to, transition)
+          update_most_recent
```
hmarr: Naming's a bit unclear here - something like `unset_old_most_recent` instead?

lib/generators/statesman/templates/create_migration.rb.erb
hmarr: If you're removing the default, there's no boolean compatibility issue here

lib/generators/statesman/templates/create_migration.rb.erb
```diff
@@ -5,9 +5,11 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
...
       t.timestamps null: false
     end
 
-    add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort  %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :most_recent], unique: true, where: "most_recent = true", name: "<%= index_name :parent_most_recent %>"
```
hmarr: Using `where: "most_recent"` should avoid the boolean literal compatibility issue here.

lib/generators/statesman/templates/update_migration.rb.erb
```diff
@@ -6,7 +6,9 @@ class AddStatesmanTo<%= migration_class_name %> < ActiveRecord::Migration
     add_column :<%= table_name %>, :<%= parent_id %>, :integer, null: false
     add_column :<%= table_name %>, :created_at, :datetime, null: false
     add_column :<%= table_name %>, :updated_at, :datetime, null: false
+    add_column :<%= table_name %>, :most_recent, :string, limit: 1, default: 't', null: false
```
hmarr: As above, `string` to `boolean`, and remove the default.

lib/generators/statesman/templates/update_migration.rb.erb
```diff
@@ -6,7 +6,9 @@ class AddStatesmanTo<%= migration_class_name %> < ActiveRecord::Migration
...
 
     add_index :<%= table_name %>, [:<%= parent_id %>, :sort_key], unique: true, name: "<%= index_name :parent_sort  %>"
+    add_index :<%= table_name %>, [:<%= parent_id %>, :most_recent], unique: true, where: "most_recent = true", name: "<%= index_name :parent_most_recent %>"
```
hmarr: Again, remove `= true` from the `where` clause.

lib/statesman/adapters/active_record.rb
```diff
@@ -72,6 +73,11 @@ def transitions_for_parent
         @parent_model.send(@transition_class.table_name)
       end
 
+      def update_most_recent
+        return unless transition_class.columns_hash["most_recent"].present?
```
hmarr: `columns_hash["most_recent"].present?` -> `columns_hash.include?("most_recent")`

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -9,22 +9,48 @@ module ClassMethods
         def in_state(*states)
           states = states.flatten.map(&:to_s)
 
-          joins(transition1_join)
-            .joins(transition2_join)
-            .where(state_inclusion_where(states), states)
-            .where("transition2.id" => nil)
+          if transition_class.columns_hash["most_recent"].present?
```
hmarr: As above

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -9,22 +9,48 @@ module ClassMethods
...
+            in_state_with_most_recent(states)
+          else
+            in_state_without_most_recent(states)
+          end
         end
 
         def not_in_state(*states)
           states = states.flatten.map(&:to_s)
 
+          if transition_class.columns_hash["most_recent"].present?
```
hmarr: And again

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -55,13 +81,19 @@ def transition2_join
              AND transition2.sort_key > transition1.sort_key"
         end
 
-        def state_inclusion_where(states)
+        def most_recent_transition_join
+          "LEFT OUTER JOIN #{transition_name} AS last_transition
+             ON #{table_name}.id = last_transition.#{model_foreign_key}
+             AND last_transition.most_recent = 't'"
```
hmarr: Removing `= 't'` should work here too

hmarr:
Generally looks good.

- If we merge, we need to make it super clear that simply creating transition *won't* do the right thing - you should always use Statesman's transition methods, and *never* modify the transition table manually, unless you really know what you're doing.
- It would also be good to verify that MySQL works properly with all this (and uses the right indexes)
- `active_record_queries.rb` is becoming unwieldily - I wonder if moving to arel would help with that. Not one for this PR, would be good to look at it soon.

lib/generators/statesman/templates/create_migration.rb.erb
petehamilton: If you leave the default, use `true` as it'll work with other SQL backends

greysteil:
@hmarr - re-review?

petehamilton:
This looks :+1: to me. @greysteil are you happy to squash and merge once the tests pass?

PR #137, opened 2015-03-16
Upgrade path for `most_recent` column

petehamilton:
Adds two migrations and a rake task to help with adding most_recent column to an existing statesman transition model.

The first migration is generated with `rails g statesman:add_most_recent`. It adds the most_recent column which will be used for writes when transitioning.

The rake task can be run with `rake statesman:backfill_most_recent[ModelName]` and backfills the most_recent column.

The final migration is generated with `rails g statesman:add_constraints_to_most_recent`. It adds a not-null constraint and a unique partial index which ensure each transition sequence can only have one "most recent" transition.

@hmarr Work in progress but would be good to get your eyes on it?

@isaacseymour FYI

### DIFF
--- a/lib/generators/statesman/templates/add_most_recent_migration.rb.erb
+++ b/lib/generators/statesman/templates/add_most_recent_migration.rb.erb
@@ -4,6 +4,6 @@ class AddMostRecentTo<%= migration_class_name %> < ActiveRecord::Migration
   end
 
   def down
-    drop_column :<%= table_name %>, :most_recent
+    remove_column :<%= table_name %>, :most_recent
   end
 end
--- a/lib/generators/statesman/templates/add_most_recent_migration.rb.erb
+++ b/lib/generators/statesman/templates/add_most_recent_migration.rb.erb
@@ -1,0 +1,9 @@
+class AddMostRecentTo<%= migration_class_name %> < ActiveRecord::Migration
+  def up
+    add_column :<%= table_name %>, :most_recent, :boolean, null: true
+  end
+
+  def down
+    drop_column :<%= table_name %>, :most_recent
+  end
+end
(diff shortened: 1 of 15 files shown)


lib/generators/statesman/templates/add_most_recent_migration.rb.erb
```diff
@@ -0,0 +1,87 @@
...
+    # Drop triggers
+    execute("DROP TRIGGER <%= table_name %>_before ON <%= table_name %>s")
+    execute("DROP FUNCTION new_<%= table_name %>_before()")
+
+    drop_column :<%= table_name %>, :most_recent
+  end
+
+  private
+
+  def add_most_recent_triggers
```
isaacseymour: :(

petehamilton: They're temporary!

greysteil:
@petehamilton - added a generator for a rake task, as people *might* want to edit its content. I'm not convinced they'd ever need to though - probably nicer to use a Railtie and make this a Statesman rake task.

greysteil:
Added a Statesman rake task too, for good measure. Let's pick an approach tomorrow.

lib/tasks/statesman.rake
```diff
@@ -0,0 +1,73 @@
...
+    total_models = parent_class.count
+    done_models = 0
+    batch_size = 500
+
+    parent_class.find_in_batches(batch_size: batch_size) do |models|
+      # Set historic transitions to have most_recent FALSE
+      #
+      # This selects transitions (initial_t) with a subsequent transition
+      # (subsequent_t), ensuring uniqueness by disallowing intermediate
+      # transitions (intermediate_t), and updates most_recent to be false.
```
Sinjo: "The LEFT JOIN eliminates duplicate entries from the first join, for example, in:

initial_t.sort_key = 10, subsequent_t.sort_key = 20
initial_t.sort_key = 10, subsequent_t.sort_key = 30

it would eliminate the second row."

hmarr:
@greysteil is confident that this all works

PR #139, opened 2015-03-18
Use the new container infrastructure on Travis

joshk:
More cpu cores, more ram, faster boot times


isaacseymour:
:+1: Thanks @joshk!

PR #140, opened 2015-03-18
Unset most_recent *before* invoking before hooks

hmarr:
Updating the parent model in a before hook causes ActiveRecord to save the parent model's associations (which includes the newly-built transition). This means the transition gets saved before we've had a chance to unmark the old most-recent transition.

This fixes that.


hmarr:
@greysteil review please

greysteil:
:+1:

PR #141, opened 2015-03-19
Add Ruby 2.2 to build matrix

greysteil:
Should really have had this since Christmas.

https://www.ruby-lang.org/en/news/2014/12/25/ruby-2-2-0-released/


greysteil:
@isaacseymour - quick one.

isaacseymour:
Why the new dev dependency? :+1: on 2.2!

greysteil:
Rails requires it and doesn't include it by default (presumably 5.0 will).

PR #145, opened 2015-03-24
Add support for JSONB metadata columns

### LINKED ISSUE #144: metadata not support jsonb format in pg 9.4.1
```ruby
t.jsonb :metadata, default: "{}"
```
will cause state machine could not be instanced.

```sh
 Statesman::UnserializedMetadataError:
 OrderTransition#metadata is not serialized
```

isaacseymour:
Closes #144 

@jackfranklin could you review?

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -6,13 +6,15 @@ module Statesman
       attr_reader :transition_class
       attr_reader :parent_model
 
+      JSON_COLUMN_TYPES = %w(json jsonb).freeze
+
       def initialize(transition_class, parent_model, observer)
         serialized = serialized?(transition_class)
         column_type = transition_class.columns_hash['metadata'].sql_type
-        if !serialized && column_type != 'json'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,6 +5,6 @@ gemspec
 gem "rails", "~> #{ENV["RAILS_VERSION"]}" if ENV["RAILS_VERSION"]
 
 # test/unit is no longer bundled with Ruby 2.2, but required by Rails
-if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.2.0")
+if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new("2.2.0")
```
greysteil: What's with the `dup`?

isaacseymour: I was getting an error where `Gem::Version` would attempt to `strip!` the version string otherwise. Was at the same time as having trouble with my 1.9.3 build though so probably related to that.

greysteil:
:+1:

isaacseymour:
Pushed as v1.2.1. Thanks again for the reminder @zw963 :dancers:

Issue #147, opened 2015-03-26
Rails 4.1.10 state_machine.current_state


romanoff:
I have noticed problem after upgrading from rails 4.1.9 to 4.1.10. `state_machine.current_state` doesn't return latest transition state anymore. Had to replace it with `self.state_machine.history.last.try(:to_state) || self.state_machine.current_state)`. After that it worked exactly the way it was supposed to. Can you please have a look at it? (I use `Statesman::Adapters::ActiveRecord` storage adapter)

tabeth:
I can confirm that this issue persists in 4.2.4. I fixed this using a similar fix mentioned by @romanoff.

greysteil:
@romanoff / @tabeth - do you have a minimal example of the above? A failing spec or steps to reproduce would be perfect.

hmac:
Closing due to inactivity.

PR #148, opened 2015-03-26
Use latest patch versions of Rails

isaacseymour:


Sinjo:
I think [use of `"~>"](https://github.com/gocardless/statesman/blob/master/Gemfile#L5)` means we already do.

isaacseymour:
Hurm, was hoping this might expose the actual issue in #147 but tests appear to all be green.

greysteil:
No harm in merging this anyway, though.

PR #149, opened 2015-03-27
Implement Allowed events

j-manu:
`allowed_events` is the events counterpart of allowed_transitions. Like `allowed_transitions`, `allowed_events` check whether the applicable guards pass

The `available_events` method do not check if the guards pass. I'm not sure whether there is a legitimate use case for this method but I have not removed it.


j-manu:
Closing this. I will create another PR which also implements "can_trigger?(event)"

PR #150, opened 2015-03-28
Implement can_trigger?(event) and allowed_events

j-manu:
`can_trigger?(event)` is the events counterpart to `can_transition_to?`

`allowed_events` is the events counterpart of `allowed_transitions`.
Like `allowed_transitions`, `allowed_events` check whether the
applicable guards pass

The `available_events` method do not check if the guards pass. I'm not sure whether there is a legitimate use case for this method and hence I have not removed it.

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -260,11 +260,23 @@ module Statesman
       false
     end
 
+    def can_trigger?(event_name)
+      state = current_state
+      transitions = self.class.events[event_name.to_sym]
+      transitions.key?(state) && can_transition_to?(transitions[state].first)
+    end
+
     def available_events
       state = current_state
       self.class.events.map do |event, transitions|
         event if transitions.key?(state)
       end.compact
+    end
+
+    def allowed_events
+      self.class.events.keys.select do |event_name|
+        can_trigger?(event_name)
+      end
     end
 
     private
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -262,9 +262,9 @@ module Statesman
 
     def available_events
       state = current_state
-      self.class.events.select do |_, transitions|
-        transitions.key?(state)
-      end.map(&:first)
+      self.class.events.map do |event, transitions|
+        event if transitions.key?(state)
+      end.compact
     end
 
     private
--- a/spec/statesman/machine_spec.rb
+++ b/spec/statesman/machine_spec.rb
@@ -749,4 +749,113 @@ describe Statesman::Machine do
       expect(instance.available_events).to eq([:event_2, :event_3])
     end
   end
+
+  describe "#can_trigger?" do
+    before do
+      machine.class_eval do
+        state :x, initial: true
+        state :y
+        state :z
+
+        event :event_1 do
+          transition from: :x, to: :y
+        end
+
+        event :event_2 do
+          transition from: :y, to: :z
+        end
+
+        event :event_3 do
+          transition from: :x, to: :z
+        end
+      end
+    end
+
+    let(:instance) { machine.new(my_model) }
+    subject { instance.can_trigger?(event) }
+
+    context "with valid event" do
+      let(:event) { :event_1 }
+      it { is_expected.to be_truthy }
+
+      context "with failing guard" do
+        let(:guard_cb) { ->(*_args) { false } }
+        let(:new_state) { :y }
+        before { machine.guard_transition(to: new_state, &guard_cb) }
+
+        it { is_expected.to be_falsey }
+      end
+
+      context "with passing guard" do
+        let(:guard_cb) { ->(*_args) { true } }
+        let(:new_state) { :y }
+        before { machine.guard_transition(to: new_state, &guard_cb) }
+
+        it { is_expected.to be_truthy }
+      end
+    end
+
+    context "with invalid event" do
+      let(:event) { :event_2 }
+      it { is_expected.to be_falsey }
+
+      context "and is guarded" do
+        let(:guard_cb) { -> { false } }
+        let(:new_state) { :z }
+        before { machine.guard_transition(to: new_state, &guard_cb) }
+
+        it "does not fire guard" do
+          expect(guard_cb).not_to receive(:call)
+          is_expected.to be_falsey
+        end
+      end
+    end
+  end
+
+  describe "#allowed_events" do
+    before do
+      machine.class_eval do
+        state :x, initial: true
+        state :y
+        state :z
+
+        event :event_1 do
+          transition from: :x, to: :y
+        end
+
+        event :event_2 do
+          transition from: :y, to: :z
+        end
+
+        event :event_3 do
+          transition from: :x, to: :z
+        end
+      end
+    end
+
+    let(:instance) { machine.new(my_model) }
+    subject { instance.allowed_events }
+
+    context "with multiple possible events" do
+      it { is_expected.to eq([:event_1, :event_3]) }
+    end
+
+    context "with one possible event" do
+      before { instance.trigger!(:event_1) }
+      it { is_expected.to eq([:event_2]) }
+    end
+
+    context "with no possible transitions" do
+      before { instance.trigger!(:event_3) }
+      it { is_expected.to eq([]) }
+    end
+
+    context "with guarded transitions" do
+      let(:guard_cb) { ->(*_args) { false } }
+      let(:new_state) { :z }
+      before { machine.guard_transition(to: new_state, &guard_cb) }
+
+      it { is_expected.to eq([:event_1]) }
+    end
+  end
 end


barisbalic:
Hi @j-manu, thanks for submitting this, internally we've been discussing events, their implementation and whether they belong in Statesman.  It seems like we're likely to be deprecating them in the near future, so we're not going to accept this PR, but we appreciate your work and your interest.  Please feel free to ask any further questions, but for now I'll close the thread.

j-manu:
Ok. But I urge you to reconsider :smiley: . I started using statesman because of the transition history and without  events it seemed simpler. But as the project grew, using events resulted in much better code. 
The problems arise when you can go back and forth between two states. 

### Before

```ruby
class OrderStateMachine
  include Statesman::Machine

  state :pending, initial: true
  state :purchased
  state :confirmed

  transition from: :pending, to: :purchased
  transition from: :purchased, to: :confirmed
  transition from: :confirmed, to: :purchased
end

```

```ruby
class Ability
  include CanCan::Ability

  def initialize(user)
    if user.customer_support?
      can :confirm, Order do |order|
        order.transition_to?(:confirmed)
      end

      can :unconfirm, Order do |order|
        order.state == 'confirmed' &&
        order.can_transition_to?(:purchased)
      end

    end
  end

```

```ruby
class OrdersController

  def confirm
    if can?(:confirm, @order)
      @order.transition_to!(:confirmed)
    end
  end

  def unconfirm
    if can?(:unconfirm, @order)
      @order.transition_to!(:purchased)
    end
  end

end

```

### After

```ruby
class OrderStateMachine
  include Statesman::Machine

  state :pending, initial: true
  state :purchased
  state :confirmed

  event :purchase do
    transition from: :pending, to: :purchased
  end

  event :confirm do
    transition from: :purchased, to: :confirmed
  end

  event :unconfirm do
    transition from: :confirmed, to: :purchased
  end
end

```

```ruby
class Ability
  include CanCan::Ability

  def initialize(user)
    if user.customer_support?
      [:confirm, :unconfirm].each do |event|
        can event, Order do |order|
          order.can_trigger?(event)
        end
      end
    end
  end
```


```ruby
class OrdersController

  def trigger_event
    event = params[:event].to_sym
    if can?(event, @order)
      @order.trigger!(event)
    end
  end
end

```

IMO, using events resulted in clearer and terser code with all the state information captured in the state machine definition itself

isaacseymour:
The main problem we have with events is that this sort of logic doesn't really live in the state machine - choosing which state you want to transition to is definitely app logic. For this use case, I think you could just use:

```ruby
class OrderStateMachine
  include Statesman::Machine

  state :pending, initial: true
  state :purchased
  state :confirmed

  transition from: :pending, to: :purchased
  transition from: :purchased, to: :confirmed
  transition from: :confirmed, to: :purchased

  def self.action_to_state(action)
    case action
    when :confirm   then :confirmed
    when :unconfirm then :purchased
    else raise "Unknown action #{action}"
    end
  end
end
```

```ruby
class Ability
  include CanCan::Ability

  def initialize(user)
    if user.customer_support?
      [:confirmed, :purchased].each do |state|
        can action, Order do |order|
          order.can_transition_to?(state)
        end
      end
    end
  end
```

```ruby
class OrdersController

  def trigger_event
    state = OrderStateMachine.action_to_state(params[:event].to_sym)
    if can?(state, @order)
      @order.transition_to!(state)
    end
  end
end
```

barisbalic:
Actually @j-manu we prescribe the opposite, we think states and rules associated with them are the concern of the state machine and the state machine alone.

I think these examples are strange because you're tying access control directly to the states of `Orders`.  What you've presented is a situation where a customer service agent, has privileges dynamically based on the states of `Orders`, and that's not typical.  The usual case is that a customer service agent is allowed to act on any given `Order`, but due to circumstance some `Orders` themselves can't be acted upon.  They are two different scenarios, which you're kind of blending into one.

I would suggest that your ACL should deal with the permissions to do things only, and your state machines should deal with the correctness of transitions only.  If you want to limit the `confirmation` of an `Order` to customer support agents, then you could use the guard clauses that Statesman provides, in conjunction with helpers or service objects.

```ruby
  guard_transition(to: :confirmed) do |order|
    user.customer_support? || user.big_boss?
  end
```

The benefit of doing this is that you reduce the coupling between your ACL and your FSM, making it easier to swap either out should you decide to.  And of course, by reducing the coupling you make things easier to test across the board.  I hope some of this is useful, it may not suit your use-case, please feel free to continue the conversation.

j-manu:
> The usual case is that a customer service agent is allowed to act on any given Order, but due to circumstance some Orders themselves can't be acted upon

That's exactly what I'm trying to do
```ruby
if user.customer_support?
  can :confirm, Order do |order|
    order.can_transition_to?(:confirmed)
  end
end
```
The above code means that customer support can confirm all orders, but orders have to be "confirmable". All the checks for whether an order is confirmable is in the statemachine as guards. I avoided guards in the example code to keep it small.

```ruby
guard_transition(to: :confirmed) do |order|
  user.customer_support? || user.big_boss?
end
```

1. `user` is dynamic and not readily available in models. Have to use Thread.current for that
2. This couples the ACL & SM together. I don't understand how it reduces the coupling between two.

I think it will be easier to reason about this at a higher level. Say a customer support person is viewing an order. We need to know what actions can be taken on the order. For simplicity, let's restrict the actions to `confirm` and `unconfirm`

#### Transitions Only, Totally decoupled ACL & SM
```ruby
guard_transition(to: :confirmed) do |order|
  order.products_in_stock?
end
```


```ruby
if user.customer_support?
  can :confirm, Order
  can :unconfirm, Order
end
```

```ruby
class Order
  def confirmable?
    can_transition_to?(:confirmed)
  end

  def unconfirmable?
    state == 'confirmed' &&
    can_transition_to?(:purchased)
  end
end
```

```erb+html

<% if can?(:confirm, @order) && @order.confirmable? %>
  <%= confirm_button %>
<% end %>

<% if can?(:unconfirm, @order) && @order.unconfirmable? %>
  <%= unconfirm_button %>
<% end %>

```

Since `purchased` can be arrived at using two separate paths,
can_transition_to?(:purchased) is not enough. If we use just that
`unconfirm` button will show up for orders in `pending` stage too.
We need to know the current state to determine the type of action.


### Transitions only, refactored

```ruby
if user.customer_support?
  can :confirm, Order do |order|
    order.confirmable?
  end
  can :unconfirm, Order do |order|
    order.unconfirmable?
  end
end
```

```ruby
class Order
  def confirmable?
    can_transition_to?(:confirmed)
  end

  def unconfirmable?
    state == 'confirmed' &&
    can_transition_to?(:purchased)
  end
end
```

```erb+html

<% if can?(:confirm, @order) %>
  <%= confirm_button %>
<% end %>

<% if can?(:unconfirm, @order)  %>
  <%= unconfirm_button %>
<% end %>

```


### Using Events

```ruby
guard_transition(to: :confirmed) do |order|
  order.products_in_stock?
end
```


```ruby
if user.customer_support?
  can :confirm, Order do |order|
    order.can_trigger?(:confirm)
  end
  can :unconfirm, Order do |order|
    order.can_trigger?(:unconfirm)
  end
end
```

```ruby
class Order

end
```

```erb+html

<% if can?(:confirm, @order) %>
  <%= confirm_button %>
<% end %>

<% if can?(:unconfirm, @order) %>
  <%= unconfirm_button %>
<% end %>

```

barisbalic:
Your totally decoupled example makes the most sense to me, I think it reads best.  The issue you have is that `unconfirmed` is not a state, but you want behaviour around it as though it is, and of course that means you need to consult the state machine further.  You could introduce it as a state, it may be useful, but I can see reasons why you wouldn't want to.

Assuming that you don't add `unconfirmed` as a state, it seems more like your problem is a presentation concern.  Nothing about the states is incorrect, but options you want to present differ from allowed transitions.  You could consider using some presentation pattern to clean this up.

Another alternative is to keep the `unconfirmable?` and `confirmable?` checks in the state machine, and delegate to the state machine from the model.

j-manu:
> Nothing about the states is incorrect, but options you want to present differ from allowed transitions

The transitions are allowed. It is just that you need to consult the state machine to figure out the exact transitions.

Yes, there are multiple ways to approach it but I think the events approach is the cleanest. With events I was able to remove the "confirmable/unconfirmable" checks and the code is more coherent. The example I gave is simplified but in my case there were 5/6 such checks each for 4 models. The reason being that it should be possible to "undo" some transactions like confirm/unconfirm.

Using events resulted in a good hierarchy with lesser code.

```
can?(:action) in View -> ACL check -> SM check -> Transition check -> Guard check
```

To each his own :smiley:  For now I have monkey patched statesman. But I guess I will have to replace it eventually.

appleton:
I think this might be a good time to consider a separate plugin. @j-manu I bet you could factor all of the events code into a separate module which you could bundle up as a gem and include in your state machine classes. What do you think?

j-manu:
I'm thinking whether that or using AASM or something similar would be the better option :)

PR #152, opened 2015-04-04
Sequel adapter

appleton:
I'm sure this needs more work, specifically there seems to be a fair bit of behaviour in the ActiveRecord adapter only which would also be applicable here, but I'm opening a PR to get some discussion going.

I'm thinking I could factor out a lot of the active record adapter specs into another shared suite, something like `it_behaves_like "a SQL adapter"`.

I guess it would also be useful to implement the queries interface.


greysteil:
This looks really nice @appleton. Agree it would be really nice to add the queries interface - would also be good to use a `most_recent` column to make that fast (like we do with ActiveRecord).

lib/statesman.rb
```diff
@@ -11,6 +11,7 @@ module Adapters
              "statesman/adapters/active_record_transition"
     autoload :ActiveRecordQueries,
              "statesman/adapters/active_record_queries"
+    autoload :Sequel, "statesman/adapters/sequel"
```
greysteil: What's going on with the whitespace here?

petehamilton:
@appleton hey hey :smile: We're thinking ahead to a 2.0 release this side of Christmas, feels like this would be a nice PR to get in? Is there anything left to do on it before it can be merged?

appleton:
I think if the tests pass then it should be good to go though we're not using this in any project yet so I can't speak to its stability. We are just considering pulling statesman into a project though so it could be good timing :)

appleton:
Well, that said there's no query interface but maybe its okay for that to come later?

petehamilton:
Sorry for the delay in getting back to you! Given this isn't a breaking change & there's no real rush on getting it in, I wonder if it's worth waiting until you're using it in production? What do you think?

appleton:
Works for me!

greysteil:
@appleton - any update? Would love to get this in if you've battle-tested it!

appleton:
It's not battle tested, we backed out of the change we were planning to make unfortunately. I'd suggest one of:

- Close this PR and forget about it for ever
- Merge it and mark it as experimental in the docs (maybe even add some feature flag in the code?)
- Extract it into its own gem. I think it actually makes more sense to say statesman is written for Active Record but adapters exist separately for other ORMs. This would also let you off the hook for maintaining the Mongo adapter why doesn't get much love :)

greysteil:
I'm for extracting into a gem - this isn't going to get the same level of support as the rest of Statesman, and I don't want it to slow down development. @hmarr / @petehamilton - what do you reckon?

hmarr:
My preference would also be extracting into a gem. Seems like a sensible approach for alternative adapters in general.

badosu:
@appleton I am looking for a state machine gem and I am very in favour of this one. However tha lack of Sequel support is weighting in.

Apart from the query interface, is this adapter working as it should with statesman? If so, I can fork your adapter and help extracting it into a gem.

appleton:
Hey @badosu yeah it should be working just fine, it passes the adapter tests at the very least. I'd be happy to help out with advice or possibly contributions if you want to take it on, I think it's a great idea!

badosu:
@appleton Thanks for the quick response.

We are going to implement a state machine on our system in the next weeks, I will update you when we come back to this.

lib/statesman/adapters/sequel.rb
```diff
@@ -0,0 +1,81 @@
+require_relative "../exceptions"
+
+module Statesman
+  module Adapters
+    class Sequel
+      attr_reader :transition_class, :parent_model, :observer
+
+      ::Sequel.extension(:inflector)
+
+      def initialize(transition_class, parent_model, observer)
```
badosu: This method now requires an additional `options={}` argument to work on the latest version.

lib/statesman/adapters/sequel.rb
```diff
@@ -0,0 +1,81 @@
...
+        create_transition(from.to_s, to.to_s, metadata)
+      ensure
+        @last_transition = nil
+      end
+
+      def history
+        history_dataset.all
+      end
+
+      def last
```
badosu: This method now accepts a `force_reload` keyword argument, this code can be refactored to support Ruby < 2.0.

```ruby
def last(force_reload: false)
  @last_transition = nil if force_reload
  @last_transition ||= history_dataset.last
end
```

badosu:
@appleton I will start using your code on the next few days, I'll post my findings and refine the adapter as it will be battle tested on the next weeks.

badosu:
@greysteil @hmarr Is there any reason for preferring using null values on the `most_recent` column when nullable? https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L100

I am developing this specific part for the adapter and this adds considerable complexity, I am not sure of the benefit. Thanks!

greysteil:
It was a requirement for people adding a `most_recent` column to an existing state machine - i.e., upgrading from Statesman 1.1 to 1.2 (version numbers may be wrong).

Given that the sequel adapter is new, there's no need to support that upgrade path, and you don't need to worry about the `most_recent` column ever being nullable.

badosu:
@greysteil Great! Thanks for the follow up.

badosu:
@greysteil @appleton We finally can attest for sure that the code is running fine on production fo rmonths and feel confident to contribute the changes.

We can update this PR or create a gem for this: `sequel-statesman` or `statesman-sequel`.

@appleton Let me know how it works best for you, since if we publish this in a separate repo we'll lose your commit history, but I'd like to make certain that your contribution credit is not lost.

appleton:
Hey @badosu, glad to hear it's working! Please do take it and create your own project, I'm fine with losing the attribution :)

greysteil:
@hmarr can you or someone from GC make the call on PR versus separate gem? Latter makes sense to me but I'm the world's worst maintainer whilst I'm on the road.

hmarr:
I'd rather go with a separate gem so there's clear ownership. Having the Mongo adapter in the main repo can make it harder to work on Statesman, as most of the core contributors aren't that familiar with Mongo. Moving future adapters to separate gems seems like a sensible approach.

Thanks for pushing this forward @badosu!

badosu:
Thanks @hmarr, @greysteil, @appleton!

Packaging it into a gem will also help in regard of an included optional sequel plugin. That would require a separate gem from statesman anyways.

I'll report back as soon as I finish polishing the gem.

badosu:
@appleton, @hmarr, @appleton 

I just released a very preliminar version at https://github.com/badosu/statesman-sequel

I'll consider it ready for use when I have migrated all the adapter specs to my suite. For now I am testing the basic behaviour with the plugin included in the gem.

Thanks for everyone!

hmac:
As of https://github.com/gocardless/statesman/pull/250 we'll link to that in the README - thanks!

PR #158, opened 2015-04-12
Fix MySQL most_recent column

### LINKED ISSUE #157: [MySQL] Uniqueness constraint on most recent index
Hi,

Perhaps I'm missing something but I'm struggling to use more than 3 states in light of the uniqueness constraint on the most recent index.  As I understand it, the index key is constructed from the id of the model and the boolean flag from the most_recent column.  My model relates to an absence, and I have the following states:

Unapproved (default state)
Provisionally Approved (sets most_recent to 1 giving key id-1)
Approved (sets most_recent to 1 giving 2 keys: id-0 for the original transition and id-1 for the transition to approved)
Cancelled - fails on transition here because a duplicate index key is created.

Is this the intended behaviour?  For the time being I've dropped the uniqueness constraint on the index but I understand this will then stop the most_recent column being used for queries.

Thanks

Robin

greysteil:
Should fix #157 by using a unique index with `NULL` values (which SQL treats as distinct from each other) rather than a partial index (which some databases don't support).


greysteil:
@isaacseymour - can you review?

greysteil:
Just discovered `ActiveRecord::Base.connection.supports_partial_index?`. Sometimes, Rails, I could kiss you ;)

spec/support/active_record.rb
```diff
@@ -50,26 +50,48 @@ def change
...
+    else
+      add_index :my_active_record_model_transitions,
+                [:my_active_record_model_id, :most_recent],
+                unique: true,
+                name: "index_my_active_record_model_transitions_"\
+                      "parent_most_recent"
+    end
+  end
+
+  def database_supports_partial_index?
```
isaacseymour: Could this just refer to the definition in the ActiveRecord adapter instead?

isaacseymour:
Looks good! :+1:

greysteil:
@isaacseymour - re-review? I'm happy with this now.

lib/tasks/statesman.rake
```diff
@@ -15,9 +15,14 @@ namespace :statesman do
 
     parent_class.find_in_batches(batch_size: batch_size) do |models|
       ActiveRecord::Base.transaction do
-        # Set all transitions' most_recent to FALSE
-        transition_class.where(parent_fk => models.map(&:id)).
-          update_all(most_recent: false)
+        if Statesman::Adapters::ActiveRecord.database_supports_partial_indexes?
```
Sinjo: Should this conditional use the column nullability too?

spec/support/active_record.rb
```diff
@@ -50,26 +50,40 @@ def change
       t.string  :to_state
       t.integer :my_active_record_model_id
       t.integer :sort_key
-      t.boolean :most_recent, default: true, null: false
 
-      # MySQL doesn't allow default values on text fields
```
Sinjo: Was this comment deliberately deleted?

Sinjo:
:+1: with minor comments.

Nice one getting this working with MySQL!

PR #166, opened 2015-06-16
Make specs run with RSpec 3.3

isaacseymour:
rspec-rails 3.3 seems to need all of Rails to be required in manually or it tries to use undefined constants all over the place.


barisbalic:
Rails: destroyer of hope, eraser of choices. :+1: if it greenz

PR #169, opened 2015-06-17
Rake backfill_most_recent task is now db-agnostic.

### LINKED ISSUE #168: Rake backfill_most_recent doesn't work in MySQL
More specifically, UPDATE …. FROM works fine in PostgreSQL, but is non-ANSI and unsupported by MySQL.

timothyp:
By virtue of using AREL instead of db-specific (PostgreSQL) SQL.

Fixes #168

### DIFF
--- a/lib/tasks/statesman.rake
+++ b/lib/tasks/statesman.rake
@@ -25,24 +25,26 @@ namespace :statesman do
         end
 
         # Set current transition's most_recent to TRUE
-        ActiveRecord::Base.connection.execute %{
-          UPDATE #{transition_class.table_name}
-          SET most_recent = true
-          FROM
-          (
-            SELECT initial_t.id, subsequent_t.created_at
-            FROM #{transition_class.table_name} initial_t
-            LEFT JOIN #{transition_class.table_name} subsequent_t ON
-            (
-                initial_t.#{parent_fk} = subsequent_t.#{parent_fk}
-                AND initial_t.sort_key < subsequent_t.sort_key
-            )
-            WHERE initial_t.#{parent_fk}
-              IN (#{models.map { |p| "'#{p.id}'" }.join(',')})
-            AND subsequent_t.id IS NULL
-          ) x
-          WHERE #{transition_class.table_name}.id = x.id
-        }
+        initial_t = transition_class.arel_table
+        subsequent_t = initial_t.alias
+
+        later_row_for_same_parent = initial_t[parent_fk].
+                                    eq(subsequent_t[parent_fk]).
+                                    and(initial_t[:sort_key].
+                                    lt(subsequent_t[:sort_key]))
+
+        no_later_row = subsequent_t[:id].eq(nil)
+        in_current_parent_batch = initial_t[parent_fk].in(models.map(&:id))
+
+        latest_ids_query = initial_t.join(subsequent_t, Arel::Nodes::OuterJoin).
+                           on(later_row_for_same_parent).
+                           where(no_later_row.and(in_current_parent_batch)).
+                           project(initial_t[:id]).to_sql
+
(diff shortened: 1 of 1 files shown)


lib/tasks/statesman.rake
```diff
@@ -25,24 +25,26 @@ namespace :statesman do
...
+        no_later_row = subsequent_t[:id].eq(nil)
+        in_current_parent_batch = initial_t[parent_fk].in(models.map(&:id))
+
+        latest_ids_query = initial_t.join(subsequent_t, Arel::Nodes::OuterJoin).
+                           on(later_row_for_same_parent).
+                           where(no_later_row.and(in_current_parent_batch)).
+                           project(initial_t[:id]).to_sql
+
+        latest_ids = transition_class.find_by_sql(latest_ids_query).
+                     to_a.collect(&:id)
```
isaacseymour: (I don't know much about arel but...) is a way to do this with a `SELECT id FROM...`? If I understand correctly, this will load up all the fields of the transition table into memory and then loop over that to pull of the IDs.

isaacseymour:
This looks sweet! I'm :+1: as-is (but would like a :+1: from @petehamilton or @greysteil who wrote the original job together).

It would be nice to add some batching to the task though - if it's operating on a very large transition table it'll run some very slow `SELECT` queries, then eat a whole load of RAM with the IDs in memory, then run another slow `UPDATE` - batching this into chunks would make the slow queries less slow, and the memory usage pretty skinny. I should probably do this myself in a separate PR though :P

lib/tasks/statesman.rake
timothyp: I'm no AREL guru either :)

The project() means we only retrieve the id.  The query that pops out is:

    SELECT `transitions`.`id` FROM `transitions`
      LEFT OUTER JOIN `transitions` `transitions_2`
        ON `transitions`.`parent_id` = `transitions_2`.`parent_id`
          AND `transitions`.`sort_key` < `transitions_2`.`sort_key`
      WHERE `transitions_2`.`id` IS NULL
        AND `transitions`.`parent_id` IN (the big list of parent ids)

isaacseymour: Sweet :+1:

timothyp:
The batching is still as per the original code; i.e. still done before this chunk of code by batching the parent models, so effectively the transition queries are still batched, just not directly if that makes sense.

isaacseymour:
Yeah, sorry - I originally just read the changes, and missed the wider context.

petehamilton:
Bit frustrated with my past-self that we didn't spec this, but after manually inspecting the AREL it looks good to me - :+1: 

@greysteil do you want to give this an eyeball as well?

timothyp:
I must say I felt bad changing code without a test.  I wondered about this in the original gitter chat a couple of months back.  It is a one-off task, but will be run against prod. databases...  On the plus side, it's behaved on my dev/staging/prod, and the constraints on the most_recent field give some protection against breakage.  I did manually check a bunch of rows and the overall table for sanity though! (i.e. checking that "select count(distinct parent_id)) from transitions" equals "select count(*) from transitions where most_recent=1" and checking that the last transition row for a parent is the one which is set to 1 (for a few different parents).

I think testing could be added either by testing the rake task (not familiar with how to do this though), or pulling the body out into a plain class for easy unit testing.

isaacseymour:
Yeah I totally agree - specs on this would be nice

greysteil:
:+1:. Agree specs would be a great improvement, but since I don't think this code will be edited again (famous last words) I'm happy as-is.

greysteil:
Thanks @timothyp - I've pushed 1.2.5 to rubygems.

timothyp:
Excellent, thanks!

PR #173, opened 2015-06-23
Remove whitespace characters in generator

samward:
When running the generator as described in your readme under Rails 4, the resulting file has a line with just two spaces on line 4.


greysteil:
:+1: been meaning to clean this up for a while!

PR #176, opened 2015-06-30
Use table_name instead of transition_name

0tsuki:
Thank you guys for this great gem. I use this on some production apps:)

I would like to fix this bug.

## Error case
https://github.com/gocardless/statesman#model-scopes

> If the transition class-name differs from the association name, you will also need to define a corresponding transition_name class method:

```ruby
class Order < ActiveRecord::Base
  has_many :transitions, class_name: "OrderTransition"

  private

  def self.transition_name
    :transitions
  end

  def self.transition_class
    OrderTransition
  end

  def self.initial_state
    OrderStateMachine.initial_state
  end
end
```

But...
```
[9] pry(main)> Order.in_state(:pending)
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'myapp_development.transitions' doesn't exist: SHOW KEYS FROM `transitions`
from /Users/otsuki.shinsuke/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-4.2.2/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:299:in `query'
[10] pry(main)> Order.send(:use_most_recent_column?)
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'myapp_development.transitions' doesn't exist: SHOW KEYS FROM `transitions`
from /Users/otsuki.shinsuke/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-4.2.2/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:299:in `query'
```

### DIFF
--- a/spec/statesman/adapters/active_record_queries_spec.rb
+++ b/spec/statesman/adapters/active_record_queries_spec.rb
@@ -15,7 +15,14 @@ describe Statesman::Adapters::ActiveRecordQueries, active_record: true do
 
   before do
     MyActiveRecordModel.send(:include, Statesman::Adapters::ActiveRecordQueries)
+    MyActiveRecordModel.send(:has_many,
+                             :transitions,
+                             class_name: 'MyActiveRecordModelTransition')
     MyActiveRecordModel.class_eval do
+      def self.transition_name
+        :transitions
+      end
+
       def self.transition_class
         MyActiveRecordModelTransition
       end
@@ -68,7 +75,7 @@ describe Statesman::Adapters::ActiveRecordQueries, active_record: true do
 
         it { is_expected.to include model }
         it { is_expected.not_to include other_model }
-        its(:to_sql) { is_expected.to include('most_recent') }
+        its(:to_sql) { is_expected.to include('most_recent =') }
       end
 
       context "given multiple states" do
@@ -108,7 +115,7 @@ describe Statesman::Adapters::ActiveRecordQueries, active_record: true do
         subject { MyActiveRecordModel.not_in_state(:failed) }
         it { is_expected.to include model }
         it { is_expected.not_to include other_model }
-        its(:to_sql) { is_expected.to include('most_recent') }
+        its(:to_sql) { is_expected.to include('most_recent =') }
       end
 
       context "given multiple states" do
@@ -137,6 +144,7 @@ describe Statesman::Adapters::ActiveRecordQueries, active_record: true do
         subject { MyActiveRecordModel.in_state(:succeeded) }
 
         it { is_expected.to include model }
+        its(:to_sql) { is_expected.not_to include('most_recent =') }
       end
 
       context "given multiple states" do
@@ -166,6 +174,7 @@ describe Statesman::Adapters::ActiveRecordQueries, active_record: true do
         subject { MyActiveRecordModel.not_in_state(:failed) }
         it { is_expected.to include model }
         it { is_expected.not_to include other_model }
+        its(:to_sql) { is_expected.not_to include('most_recent =') }
       end
 
       context "given multiple states" do
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -127,10 +127,10 @@ module Statesman
         # it has good data
         def use_most_recent_column?
           ::ActiveRecord::Base.connection.index_exists?(
-            transition_name,
+            transition_class.table_name,
             [model_foreign_key, :most_recent],
             unique: true,
-            name: "index_#{transition_name}_parent_most_recent"
+            name: "index_#{transition_class.table_name}_parent_most_recent"
           )
         end
       end


greysteil:
:+1: good fix. Can you write a spec, too?

0tsuki:
Yes. I got it.

0tsuki:
@greysteil re-review?

spec/statesman/adapters/active_record_queries_spec.rb
```diff
@@ -68,7 +75,7 @@ def self.initial_state
 
         it { is_expected.to include model }
         it { is_expected.not_to include other_model }
-        its(:to_sql) { is_expected.to include('most_recent') }
+        its(:to_sql) { is_expected.to include('most_recent =') }
```
greysteil: Any reason for changing these? I think I prefer them as they were - the SQL query could be `WHERE most_recent` rather than `WHERE most_recent = 't'`, so I'd rather not add the `=`.

spec/statesman/adapters/active_record_queries_spec.rb
```diff
@@ -137,6 +144,7 @@ def self.initial_state
         subject { MyActiveRecordModel.in_state(:succeeded) }
 
         it { is_expected.to include model }
+        its(:to_sql) { is_expected.not_to include('most_recent =') }
```
greysteil: Lose the `=` as above, but I'm :+1: on the additional specs.

spec/statesman/adapters/active_record_queries_spec.rb
```diff
@@ -15,7 +15,14 @@
 
   before do
     MyActiveRecordModel.send(:include, Statesman::Adapters::ActiveRecordQueries)
+    MyActiveRecordModel.send(:has_many,
```
greysteil: :+1: on the change, but I'd do this is a separate `context` block. That way it will be more explicit that we're testing this case in particular and no overzealous refactoring will change these specs :)

greysteil:
Looks awesome - thanks for fixing this! If you give this PR 5 mins cleanup I'll get it merged and released today. :octocat:

spec/statesman/adapters/active_record_queries_spec.rb
0tsuki: I thought so too. but I realized that this spec returned true at all times.
```
$ git checkout 7fd6989 # my commit
$ bundle exec rake
/Users/otsuki.shinsuke/.rbenv/versions/2.2.1/bin/ruby -I/Users/otsuki.shinsuke/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/rspec-core-3.3.1/lib:/Users/otsuki.shinsuke/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/rspec-support-3.3.0/lib /Users/otsuki.shinsuke/.rbenv/v
ersions/2.2.1/lib/ruby/gems/2.2.0/gems/rspec-core-3.3.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
Running with database adapter 'SQLite'

Randomized with seed 23601
................................................................................................................................................................................................................................................................................
.......F....F.............

Failures:

  1) Statesman::Adapters::ActiveRecordQueries with a most_recent column .in_state given a single state to_sql should include "most_recent ="
     Failure/Error: its(:to_sql) { is_expected.to include('most_recent =') }
       expected "SELECT \"my_active_record_models\".* FROM \"my_active_record_models\" LEFT OUTER JOIN my_active_record_model_transitions most_recent_transition\n             ON most_recent_transition.my_active_record_model_id =\n                  my_active_record_models.
id LEFT OUTER JOIN my_active_record_model_transitions other_transition\n             ON other_transition.my_active_record_model_id =\n                  my_active_record_models.id\n             AND other_transition.sort_key >\n                   most_recent_transition.sort
_key WHERE (most_recent_transition.to_state IN ('succeeded') AND most_recent_transition.to_state IS NOT NULL) AND \"other_transition\".\"id\" IS NULL" to include "most_recent ="
       Diff:
       @@ -1,2 +1,8 @@
       -most_recent =
       +SELECT "my_active_record_models".* FROM "my_active_record_models" LEFT OUTER JOIN my_active_record_model_transitions most_recent_transition
       +             ON most_recent_transition.my_active_record_model_id =
       +                  my_active_record_models.id LEFT OUTER JOIN my_active_record_model_transitions other_transition
       +             ON other_transition.my_active_record_model_id =
       +                  my_active_record_models.id
       +             AND other_transition.sort_key >
       +                   most_recent_transition.sort_key WHERE (most_recent_transition.to_state IN ('succeeded') AND most_recent_transition.to_state IS NOT NULL) AND "other_transition"."id" IS NULL
     # ./spec/statesman/adapters/active_record_queries_spec.rb:78:in `block (5 levels) in <top (required)>'
```
So `most_recent` is not only column name but also table alias name, `most_recent_transition` is included `most_recent`.
However this changing is strange. any idea?

spec/statesman/adapters/active_record_queries_spec.rb
0tsuki: It makes sense to me:) but I don't know/understand what I should do in particular. I'm sorry.

spec/statesman/adapters/active_record_queries_spec.rb
greysteil: Good shout - looks like this was a bad spec. I'll update to make it `'.most_recent'`.

spec/statesman/adapters/active_record_queries_spec.rb
greysteil: No worries, I'll handle it.

greysteil:
Merged as https://github.com/gocardless/statesman/commit/a8d017bc8029b63257a7620c45f4e2cc089a5b85

greysteil:
Released in v1.3.1

0tsuki:
Thank you very much.

PR #177, opened 2015-07-01
Condition most_recent rake task on database supporting partial index

greysteil:
We only ever want to set `false` values if the database support partial indexes. Otherwise we're going to need to set `nil` values, and use the fact that SQL treats each `null` values as unique.

### DIFF
--- a/lib/tasks/statesman.rake
+++ b/lib/tasks/statesman.rake
@@ -15,7 +15,7 @@ namespace :statesman do
 
     parent_class.find_in_batches(batch_size: batch_size) do |models|
       ActiveRecord::Base.transaction do
-        if transition_class.columns_hash['most_recent'].null == false
+        if Statesman::Adapters::ActiveRecord.database_supports_partial_indexes?
           # Set all transitions' most_recent to FALSE
           transition_class.where(parent_fk => models.map(&:id)).
             update_all(most_recent: false)


greysteil:
@davegudge - can you check if this fixes your issue?

davegudge:
This appears to fix the rake task. The `most_recent` column is now set to `false` when using PostgreSQL 9.1.15.

I'm still unable to apply the constraint on the transitions table:

```
== 20150701192629 AddConstraintsToMostRecentForAllocationTransitions: migrating
-- add_index(:allocation_transitions, [:allocation_id, :most_recent], {:unique=>true, :where=>"most_recent", :name=>"index_allocation_transitions_parent_most_recent", :algorithm=>:concurrently})
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

PG::UniqueViolation: ERROR:  could not create unique index "index_allocation_transitions_parent_most_recent"
DETAIL:  Key (allocation_id, most_recent)=(14814, f) is duplicated.
: CREATE UNIQUE INDEX "index_allocation_transitions_parent_most_recent" ON "allocation_transitions" ("allocation_id", "most_recent")
```

It does not seem to be taking the `where` condition into consideration given the generated SQL in the error. I am attempting to run this against a rails 4.0.13 app, which should support this [`add_index` syntax](http://apidock.com/rails/v4.0.2/ActiveRecord/ConnectionAdapters/SchemaStatements/add_index)

davegudge:
I have found the issue. I am using the [SchemaPlus gem](https://github.com/SchemaPlus/schema_plus/tree/1.x) and [the syntax to add a partial index](https://github.com/SchemaPlus/schema_plus/tree/1.x#indexes) is slightly different. In my migration, I have updated the `where: "most_recent"` to `conditions: "most_recent"` and the migration runs successfully, creating the partial index.

Thanks for fixing the rake task issue in this PR. I would say it's good to merge :+1:

greysteil:
Ah cool, that makes sense! Irritating that SchemaPlus silently ignores the `where` option - would have been a lot easier to debug this if it raised for unknown options. A PR for someone...

davegudge:
It may be handled better is SchemaPlus v2, I have not upgraded yet.

Should I create a PR to bump the version to v1.3.1?

greysteil:
I'll handle the version bump this afternoon / evening - I'd like to get https://github.com/gocardless/statesman/pull/176 into the new version, too.

davegudge:
No problem.

greysteil:
Released in v1.3.1

davegudge:
:+1: Thank you.

Issue #179, opened 2015-08-08
Transition model should update `most_recent` after destroy


hale:
If you destroy the latest transition, there is no 'most_recent' transition for the parent model, breaking the current_state method and scopes

```ruby
class Transition < ActiveRecord::Base
  belongs_to :parent
  after_destroy :update_most_recent 

  private

  def update_most_recent
    parent.history.last.update(most_recent: true)
  end
end
```

Would something like the above work?

greysteil:
Sounds like a good idea to me. If you add a PR for that along with a test then I'm :+1:.

greysteil:
Implemented in #180 - thanks for the suggestion @hale

PR #180, opened 2015-08-12
Update most-recent after transition destroy

greysteil:
This is a good pattern suggested in #179. Just affects the ActiveRecord transition model generator.

### DIFF
--- a/lib/generators/statesman/templates/active_record_transition_model.rb.erb
+++ b/lib/generators/statesman/templates/active_record_transition_model.rb.erb
@@ -5,4 +5,14 @@ class <%= klass %> < ActiveRecord::Base
   attr_accessible :to_state, :metadata, :sort_key
 <% end %>
   belongs_to :<%= parent_name %><%= class_name_option %>, inverse_of: :<%= table_name %>
+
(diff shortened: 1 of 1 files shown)


isaacseymour:
:+1:

hale:
Thanks for adding this.  What about `after_destroy :update_most_recent, if: :most_recent?`, so we only update when necessary?

greysteil:
We're about to 2.0 statesman to make the `most_recent` column required - see https://github.com/gocardless/statesman/pull/184. This change will go out in that version, so I think we're fine here.

hale:
I was more thinking that we only care about updating the `most_recent` column if the deleted transition is the most recent - otherwise, we can skip the callback.

With the condition the `most_recent` column is checked after every destroy. Without it,  every destroy sorts the table then checks the `most_recent` column to see if it needs to `UPDATE`. 

Maybe this doesn't matter in practice.

greysteil:
Good shout. Done in https://github.com/gocardless/statesman/commit/926f45417f59a76af9fc3f143c657f3cde0cdcc1.

PR #182, opened 2015-08-21
Unset most_recent after before transitions

greysteil:
Currently we're unsetting the `most_recent` column on all a model's transitions **before** the state machine's `before_transition` callbacks have run. That means if any query in those callbacks relies on the model being in ints previous state it will get unexpected result.

This PR changes that to instead unset the `most_recent` column **after** the state machine's `before_transition` callbacks, but before its `after_transition` ones. That's considerably more intuitive, but introduces a potential gotcha: if you attempt to save the parent model during a `before_transition` and you don't have `autosave: false` on the associations you'll get a uniqueness error.

I think that's a fair price to pay for more intuitive behaviour, and it can be mitigated by setting `autosave: false` on the transition association (and this PR updates the readme to suggest that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ end
 class Order < ActiveRecord::Base
   include Statesman::Adapters::ActiveRecordQueries
 
-  has_many :order_transitions
+  has_many :order_transitions, autosave: false
 
   def state_machine
     @state_machine ||= OrderStateMachine.new(self, transition_class: OrderTransition)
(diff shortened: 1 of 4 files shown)


greysteil:
@hmarr @isaacseymour - what do you reckon? There's a versioning issue here, too - might require a major version bump?

isaacseymour:
:+1: on the change - not sure about how to version it :/

hmarr:
Major version bump sounds right to me. Blindly updating to a version that includes this behaviour could really break things for some people!

timothyp:
Excellent! I was poised to suggest this change.  I think it also reduces the chance of db transaction deadlock or lock timeout which can occur depending on how long the `before_transition` callbacks take to execute (on MySQL at least...).

Are there any expectations for when 2.0 will be released?

isaacseymour:
@timothyp I think we'll try to get it out within a week or so - there might be some other changes that go into it, yet to be decided.

timothyp:
Great, thanks @isaacseymour.

PR #185, opened 2015-08-27
Require a most_recent column on transition tables

greysteil:
Previously we were accommodating for not having a `most_recent` column. In 2.0 we should require that column.


greysteil:
@hmarr - can you review?

hmarr:
:+1:

PR #188, opened 2015-11-11
Increase the default initial sort key to 10

petehamilton:
We use intervals of 10 between sort_keys so that we can inject
intermediate states easily later on, whilst maintaining a correct
chronological ordering.

The current sort_key default for the initial state is 0, which means
that you can't really inject a new initial state easily (without going
negative).

@hmarr can you review?


hmarr:
:+1:

PR #189, opened 2015-11-15
Add in_state? instance method

hmac:
Example usage:

```ruby
machine.in_state?(:pending)
# => true

machine.in_state?("pending")
# => true

machine.in_state?(:cancelled)
# => false

machine.in_state?(:pending, :cancelled)
# => true

machine.in_state?("some nonsense state")
# => false
```

@greysteil thoughts?

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -177,9 +177,9 @@ module Statesman
     end
 
     def initialize(object,
-                      options = {
-                        transition_class: Statesman::Adapters::MemoryTransition
-                      })
+                   options = {
+                     transition_class: Statesman::Adapters::MemoryTransition
+                   })
       @object = object
       @transition_class = options[:transition_class]
       @storage_adapter = adapter_class(@transition_class).new(
@@ -190,6 +190,10 @@ module Statesman
     def current_state
       last_action = last_transition
       last_action ? last_action.to_state : self.class.initial_state
+    end
+
+    def in_state?(*states)
+      states.flatten.any? { |state| current_state == state.to_s }
     end
 
     def allowed_transitions
--- a/spec/statesman/machine_spec.rb
+++ b/spec/statesman/machine_spec.rb
@@ -374,6 +374,61 @@ describe Statesman::Machine do
     end
   end
 
+  describe "#in_state?" do
+    before do
+      machine.class_eval do
+        state :x, initial: true
+        state :y
+        transition from: :x, to: :y
+      end
+    end
+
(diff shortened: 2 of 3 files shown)


lib/statesman/machine.rb
```diff
@@ -192,6 +192,10 @@ def current_state
       last_action ? last_action.to_state : self.class.initial_state
     end
 
+    def in_state?(state)
```
greysteil: Be nice if this could handle being passed an array or multiple arguments, too.

spec/statesman/machine_spec.rb
```diff
@@ -374,6 +374,38 @@ def after_initialize; end
...
+      end
+    end
+
+    let(:instance) { machine.new(my_model) }
+    subject { instance.in_state?(state) }
+    before { instance.transition_to!(:y) }
+
+    context "when machine is in given state" do
+      let(:state) { "y" }
+
```
greysteil: :scissors: to this whitespace (since you've only got one example in the block, and its a oneliner.

greysteil:
:+1:

lib/statesman/machine.rb
```diff
@@ -177,9 +177,9 @@ def array_to_s_or_nil(input)
     end
 
     def initialize(object,
-                      options = {
-                        transition_class: Statesman::Adapters::MemoryTransition
-                      })
+                   options = {
+                     transition_class: Statesman::Adapters::MemoryTransition
+                   })
```
hmac: Happy to remove this - wanted to clean it up while I was there.

spec/statesman/machine_spec.rb
```diff
@@ -374,6 +374,47 @@ def after_initialize; end
...
+    context "when given multiple states" do
+      context "when one of the states is the current state" do
+        subject { instance.in_state?(:x, :y) }
+        it { is_expected.to eq(true) }
+      end
+
+      context "when none of the states are the current state" do
+        subject { instance.in_state?(:x, :z) }
+        it { is_expected.to eq(false) }
+      end
```
greysteil: Add a spec for passing an array?

greysteil:
:+1: after final spec. Loving your work @hmac !

spec/statesman/machine_spec.rb
hmac: Opinions on passing an array vs multiple args? Prefer one over the other, or both?

greysteil: I think we support both for the scope, so we should probably do that here, too.

greysteil:
:tada:

PR #191, opened 2015-12-05
Drop Circle

greysteil:
That. We're not using it.


greysteil:
@hmarr - any reason not to do this?

hmarr:
:+1:

PR #193, opened 2015-12-16
Make using cache for last_transition optional

petehamilton:
@greysteil What do you think? Need to add specs for other adapters but it's a first cut.

### DIFF
--- a/lib/statesman/adapters/memory.rb
+++ b/lib/statesman/adapters/memory.rb
@@ -28,7 +28,7 @@ module Statesman
         transition
       end
 
-      def last
+      def last(*)
         @history.sort_by(&:sort_key).last
       end
 
--- a/spec/statesman/adapters/active_record_spec.rb
+++ b/spec/statesman/adapters/active_record_spec.rb
@@ -207,10 +207,30 @@ describe Statesman::Adapters::ActiveRecord, active_record: true do
         adapter.last
       end
 
-      context "and a new transition" do
+      context "after then creating a new transition" do
         before { adapter.create(:y, :z, []) }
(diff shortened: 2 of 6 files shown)


lib/statesman/adapters/memory.rb
```diff
@@ -28,7 +28,7 @@ def create(from, to, metadata = {})
         transition
       end
 
-      def last
+      def last(use_cache: true)
```
petehamilton: Not hugely keen on this. Be nice if it wasn't an unused arg...

greysteil: I think that's fine, but I'm not sure about `use_cache`. Would rather have busting the cache as the explicit action, so something like `force_reload` (although that's not perfect).

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -207,12 +207,32 @@
...
+            to receive(:my_active_record_model_transitions).never
+          expect(adapter.last.to_state).to eq("y")
+        end
+
+        context "when explitly not using the cache" do
+          it "still returns the cached value" do
+            expect(adapter.last(use_cache: false).to_state).to eq("z")
+          end
+        end
+      end
```
greysteil: Can we add a shared adapter spec, too, so that all adapters have the same method signature?

greysteil:
:+1: after naming.

lib/statesman/adapters/memory.rb
petehamilton: I agree - making 'bust' be the affirmative makes more sense to me.

PR #200, opened 2016-01-17
Remove `private` declaration from class methods in README examples

andyw8:
Marking a class method as `private` has no effect, and will cause RuboCop to generate a `Useless private access modifier` warning.


greysteil:
Good shout. Can you update this to specify that these are `private_class_method`s?

Thanks!

andyw8:
Done, thanks.

greysteil:
:+1: thanks!

PR #204, opened 2016-02-28
Ensure consistency in the migration template

kenchan0130:
Hi,

There seems to be an inconsistency in the migration template.
Parent id and the `most_recent`columns should not be unique.

I would appreciate if you check this.
Thank you.

### DIFF
--- a/lib/generators/statesman/templates/create_migration.rb.erb
+++ b/lib/generators/statesman/templates/create_migration.rb.erb
@@ -15,7 +15,6 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
               name: "<%= index_name :parent_sort %>")
     add_index(:<%= table_name %>,
               [:<%= parent_id %>, :most_recent],
-              unique: true,
               <%= "where: 'most_recent'," if database_supports_partial_indexes? %>
               name: "<%= index_name :parent_most_recent %>")
   end


greysteil:
Hi @kenchan0130,

We do want this index to be unique - there should only be one `most_recent` transition for any given instance of the state machine.

kenchan0130:
Hi, @greysteil

Thank you for your reply.
But, I still don't understand your point.
There should be a unique `true` `most_recent` item per `parent_id`, but there
will be more than one `false` `most_recent` item.

For example, if the following states
```
draft → published → trashed
```
exist,
when the state is `trashed`

id|to_state|parent_id|most_recent
---|---|---|---
1|draft|100|0
2|published|100|0
3|trashed|100|1

The index on `parent_id` and `most_recent` is not unique.

What do you think?

greysteil:
Ah, sorry, this is more nuanced than I originally suggested.

The line below this one puts a caveat in to check whether your database supports partial indexes. If it does, then this index will only apply to rows where `most_recent` is true. If it doesn't, then Statesman will be setting `most_recent` to `NULL` rather than `FALSE` for older transitions. In the SQL spec `NULL != NULL`, so the index still works.

Are you having trouble with this in production? What database are you using?

kenchan0130:
Hi @greysteil,
Sorry for the late reply.

I see what you mean.
But, I look at the SQL that has issued by ActiveRecord, 

```sql
UPDATE "parent_transitions" SET "most_recent" = 'f' WHERE "parent_transitions"."parent_id" = $1  [["parent_id", 1]]
```

So, statesman is setting `most_recent` to `false`.

### env
Database: PostgreSQL 9.4
ORM: ActiveRecord 4.2.5

kenchan0130:
I'm sorry. I solved this problem myself.
For some reason the partial index had not worked.

PR #207, opened 2016-03-22
Refactors private method #add_callback

rthbound:
- Simplifies `#add_callback` method and the methods that call it.
  - Uses required arguments to require callback_class/callback_type
    rather than using `Hash#fetch` to force an error. This obviates
    the need for `#add_callback` callers to merge k/v pairs into the
    `options` argument.

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -66,25 +66,21 @@ module Statesman
         successors[from] += to
       end
 
-      def before_transition(options = { from: nil, to: nil }, &block)
-        add_callback(
-          options.merge(callback_class: Callback, callback_type: :before),
-          &block)
+      def before_transition(options = {}, &block)
+        add_callback(callback_type: :before, callback_class: Callback,
+                     from: options[:from], to: options[:to], &block)
       end
 
-      def guard_transition(options = { from: nil, to: nil }, &block)
-        add_callback(
-          options.merge(callback_class: Guard, callback_type: :guards),
-          &block)
+      def guard_transition(options = {}, &block)
+        add_callback(callback_type: :guards, callback_class: Guard,
+                     from: options[:from], to: options[:to], &block)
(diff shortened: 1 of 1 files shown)


greysteil:
Hi @rthbound,

I'd be :+1: on this if we passed the callback class and type as keyword arguments, and still raised an explicit error if either of them was `nil`. Agree that merging into the options hash was ugly - hangover from supporting Ruby < 2.0.0.

rthbound:
I almost went with kwargs, but I wanted to use required arguments and 2.0.0
kwargs don't quite support that. I'll make the changes you mentioned and
squash shortly.

Thanks for the feedback!

rthbound:
@greysteil quick question... which do you prefer: `ArgumentError` or `InvalidCallbackError`?

greysteil:
Let's got for `ArgumentError` - we'll move to required keywords when we drop support for 2.0.0, and it would be nice to have consistency then.

rthbound:
@greysteil take a look and tell me if this is what you had in mind.  If so, I'll squash.

As an aside, I noticed that about half of the `#validate_*` methods in `Statesman::Machine::ClassMethods` are private while the other half are public -- is this as desired?

greysteil:
:+1: and no, I think all of those could be private.

greysteil:
(If you squash these I'll get them merged.)

rthbound:
@greysteil I've squashed.

I decided to wait on changing method visibility for the other `#validate_*` methods. They're currently called directly in tests, so we'd need to make some changes there. Are you okay with using `#send` in tests to hit the methods once they're private, or would you prefer to refactor the tests so they cover all these `#validate_*` methods as they'd actually be called?

greysteil:
I'm cool with using `send` for now - it would still be an improvement, and I'm happy to iterate on this one.

PR #214, opened 2016-06-29
Statesman min

mkcode:
Ripping out all of the database related code from statesman to make 'statesmin' - a minimal version of statesman.

This retains the beautiful and well tested DSL that statesman uses to create state machines. All of the db related transitions and scopes are gone. This is to allow for the custom transition logging and overall behavior that I need in my application.

`Statesman::Machine` instances are mostly the same, except that the `history` and `last_transaction` methods are now removed. The current state of an object now needs to be supplied to the `Statesman::Machine` constructor, rather than it being calculated from the transitions table.

All tests pass locally.


mkcode:
Whoops. Meant this to be against my fork! Closing.

PR #217, opened 2016-07-05
Explicit errors

mkcode:
Use an explicit `NotImplementedError` when necessary parts of a transition class (one that includes TransitionHelper) are not implemented.


mkcode:
Again! Apologies! Wish I could delete after.

PR #218, opened 2016-07-07
Link to statesmin in readme

mkcode:
I created a [custom fork of statesman](https://github.com/mkcode/statesmin) that removes the all of the database related code. I'm calling it statesmin, because it is a minimal version of statesman.

For my application, the way the Transitions table is setup will not work with the requirements I have. I need a more customized version of the transitions table. I also need multiple types of transitions in my application, with different transitions log tables. I also need multiple state machines for the same model. I also will be frequently reading the current state of records, which can be better handled than looking up via a join query on a table with more than 5 million rows (in my case).

Also worth noting, that before I decided on forking statesman, I looked at all of the about 12 or so state machine libraries for ruby and could not find any that came closer to my needs than statesman. I love that statesman has a philosophy of decoupling the state machines from the model and has an elegant, simple, and well tested DSL for defining state machines.

You can check out the [Pull Requests](https://github.com/mkcode/statesmin/pulls?q=is%3Apr+is%3Aclosed) that created this minimal version of statesman. The majority of the relevant work is done here: https://github.com/mkcode/statesmin/pull/1

I provide this PR as a 0 work option for working together. I generally try not to fork libraries if possible, because of potential community splitting and additional maintenance requirements. The ideal way to move forward in my opinion would be if we could reference my fork to create a `statesman-core` like gem that also does not have DB requirements and then add that functionality back in a separate gem for those that need it, similar to how `statesman-events` was extracted out into a separate gem.

Hope you will take the time to consider this. I outline the differences on the [top of the statesmin readme](https://github.com/mkcode/statesmin#when-to-use-statesmin-over-statesman). It's a tradeoff, where statesmin can be much more flexible at the cost of more up front work to get setup.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,12 @@ describe "some callback" do
 end
 ```
 
+## Alternative minimal version
+
+If you love everything about statesman except for the hardcoded Transitions
+model, check out [statemin](https://github.com/mkcode/statesmin) - a minimal
+version of statesman.
+
 ---
 
 GoCardless ♥ open source. If you do too, come [join us](https://gocardless.com/jobs#software-engineer).


greysteil:
Thanks for the PR  @mkcode. Linking to your gem sounds sensible to me, but I'm going to tag in @hmarr and the @gocardless/backend team as they're responsible for the longer-term vision for statesman.

Team - over to you.

hmac:
Sorry for the radio silence on this. It looks like statesmin isn't under active development anymore? If you're still maintaining it we'd be happy to link to it, but otherwise I think we can close this.

Please reopen if you disagree!

Issue #220, opened 2016-07-22
Model.in_state(:valid) cannot identify correct transition model name


mecampbellsoup:
`Statesman::Adapters::ActiveRecordQueries#transition_refelection` calls `#transition_name`. However, that method gets the table name of the transition class like so:

``` ruby
        def transition_name
          transition_class.table_name.to_sym
        end
```

Therefore, `#transition_reflection` (which needs the association name, not the table name) is returning `nil`:

``` ruby
        def transition_reflection
          reflect_on_association(transition_name)
        end
```

In my case, I have `AuthToken::Transitions` that belong to an `AuthToken`. 

The code that revealed this potential bug to me was: `AuthToken.in_state(:valid)`:

```
[2] pry(AuthToken)> transition_name
:auth_token_transitions
[3] pry(AuthToken)> AuthToken.in_state(:valid)
NoMethodError: undefined method `table_name' for nil:NilClass
from /Users/mec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/statesman-2.0.1/lib/statesman/adapters/active_record_queries.rb:50:in `model_table'
[4] pry(AuthToken)> 
```

Rails 5.0.0
Ruby 2.3.0

mecampbellsoup:
@greysteil @isaacseymour are you guys maintainers still? Thanks :)

mecampbellsoup:
Here is perhaps a better display of the bug. First I invoke `#transition_reflection` showing it starts out as `nil`. Then, I override the method to use the correct `#transition_name` which in this example is `AuthToken.has_many :transitions`:

```
    48: def model_table
 => 49:   binding.pry
    50:   transition_reflection.table_name
    51: end

[1] pry(AuthToken)> transition_reflection
nil
[2] pry(AuthToken)> transition_name
:auth_token_transitions
[3] pry(AuthToken)> def transition_name
[3] pry(AuthToken)*   :transitions
[3] pry(AuthToken)* end  
:transition_name
[4] pry(AuthToken)> transition_reflection
[...]

PR #221, opened 2016-07-26
Handle custom transition association names gracefully

greysteil:
Previously, statesman required the association with the transition class to use
the exact name of the transition class, unless a `transition_name` was specified
along with the `transition_class`.

This PR updates the way we look for the association with the transition class so
as `transition_name` no longer needs to be overridden to work with custom named
associations.

Fixes https://github.com/gocardless/statesman/issues/220.

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -38,7 +38,13 @@ module Statesman
         end
 
         def transition_reflection
-          reflect_on_association(transition_name)
+          reflect_on_all_associations(:has_many).each do |value|
+            return value if value.klass == transition_class
+          end
+
+          raise MissingTransitionAssociation,
+                "Could not find has_many association between #{self.class} " \
+                "and #{transition_class}."
         end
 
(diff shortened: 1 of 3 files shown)


petehamilton:
@greysteil nice. What do you think the correct behaviour should be if there are multiple has_many definitions for the same class? Maybe prefer transition_name if it's provided?

```
class Payment
  has_many :new_transitions, -> { where('created_at > ?', Time.current - 1.year) }, class_name: "PaymentTransition" # Would we return on this one?
  has_many :transitions, class_name: "PaymentTransition"

  def self.transition_class
    PaymentTransition
  end
```

greysteil:
We actually only use the reflection to get the relevant foreign key and model table, so having multiple `has_many` definitions for the same class shouldn't cause any problems. I'd actually like to remove the `transition_name` method entirely (it'll now only be used in one place, and shouldn't need overriding). Would you be 👍  on that?

greysteil:
@petehamilton - any reason not to merge this?

hmac:
This seems legit to me.

PR #227, opened 2016-09-24
Append a thing that is able to use jsonb to README

kenchan0130:
Hi,

I read [this post](http://nandovieira.com/using-postgresql-and-jsonb-with-ruby-on-rails).

In some cases, `jsonb` makes the rails app high performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ By default, Statesman uses `serialize` to store the metadata in JSON format.
 It is also possible to use the PostgreSQL JSON column if you are using Rails 4
 or 5. To do that
 
-* Change `metadata` column type in the transition model migration to `json`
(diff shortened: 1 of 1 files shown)


nickcampbell18:
JSONb provides great performance improvements at query-time, but I think `metadata` is not usually queried alone (e.g. `Transition.where("metadata->>'attribute' = ?", "something")`).

kenchan0130:
I think `metadata` is useful in the case of auditing.

kenchan0130:
@nickcampbell18 

Do you have any update on this PR?

hmac:
Looks good to me! I think highlighting that you can choose `json` or `jsonb` makes sense.

PR #230, opened 2016-10-27
Adds options[:association_name] parameter

svenyurgensson:
It is useful when, say, you have model wrapped in namespace.

```ruby
module Models
  class Order
     embeds_many :order_transitions, class_name: 'Models::Order::Transition'
     def state_machine
             @state_machine ||= Models::Order::StateMachine.new(self,
                                                 association_name: :order_transitions,
                                                 transition_class: Order::Transition)
     end
  end
end
```


danwakefield:
We allow specifying the name of the transition table with [`transition_name:`](https://github.com/gocardless/statesman#model-scopes)

PR #234, opened 2017-01-26
Lock nokogiri dependency on Ruby 2.0

nickcampbell18:
Nokogiri 1.7 removes support for Ruby 2.0, even though Rails 4.2 will run on this version of Ruby. This downgrades the dependency to the last 2.0-supporting version of Nokogiri.


Sinjo:
A discussion for outside of this PR: we should consider removing 2.0 support as it no longer receives security patches.

nickcampbell18:
Closed in favour of #235

PR #238, opened 2017-02-21
README: Use SVG badges

olleolleolle:
This PR makes the build badges shiny.


Sinjo:
These render so much nicer than the PNGs! Thanks for the PR. ❤️

PR #240, opened 2017-04-18
Remove double-space from generated transition models

greysteil:
Also renames a confusing helper method.


greysteil:
@petehamilton - quick tweak.

Sinjo:
Cheers @greysteil!

PR #241, opened 2017-05-05
Add #transitions_to

Dorian:
Finds the path between two steps and transition one step at a time if the path is found

- [x] Is @gocardless cool with that?
  - [ ] Yes
    - [ ] Improve the code
    - [ ] Add specs

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -234,8 +234,27 @@ module Statesman
     rescue TransitionFailedError, GuardFailedError
       false
     end
+    
+    def transitions_to(to_state, opts = {})
+      states_to_transition(current_state.to_s, to_state.to_s)[1..-1]&.each do |state|
+        transition_to(to_state, opts)
+      end
+    end
 
     private
+    
+    def states_to_transition(from, to, states = [])
+      states = states.dup + [from]
+
+      return states if states.last == to || !self.class.successors[from]
+
+      paths =
+        self.class.successors[from].map do |after|
+          states_to_transition(after, to, states)
+        end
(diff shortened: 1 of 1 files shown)


Sinjo:
Hi @Dorian, thanks for taking time to contribute to Statesman!

Could you describe the use case you see for this feature? It's the first time we've had something like this proposed, and we're always keen to understand the intended use of anything we add.

Dorian:
@sinjo the idea is to be able to move to any successive state but keeping the previous transitions.

E.g. a car is built, painted, sold, shipped and used. But then we find a car that went through the process really really fast so it's salary in use. It would be nice to mark it as used but keeping all the proper state transitions it went through. So why not just do `transitions_to(:used)` from the `:built` state :)

That's a very stupid and simple example but can be applied to a lot of state machines

Sinjo:
@Dorian Thanks for your explanation. Unfortunately, we've decided that we won't be merging this PR. There are two reasons, and the second is the more important one:

1. We don't believe this represents a majority use-case. In most cases people care about the series of transitions taken to reach a given state. Leaving it to chance which path is taken through all possible transitions is undesirable more often than not.
2. We believe very strongly in code that is explicit, rather than implicit in how it works. This was the reason we moved events out from this repo and into [gocardless/statesman-events](https://github.com/gocardless/statesman-events). We feel that `transitions_to` is more implicit in what it will do than we would like from a feature in Statesman core.

Again, thanks for taking time to send us a PR.

Dorian:
👍 Might make in a gem/add-on at some point if I see some interest.

PR #243, opened 2017-05-14
Support rails5 for templates

kenchan0130:
## Topics

- support rails5 for transition model template 
  - see also: https://github.com/rails/rails/pull/22567
- support rails5 for generator templates
  - see also: https://github.com/rails/rails/pull/21538

## TODO
- [x] fix test
  - Since the `DEPRECATION WARNING` has been put in rails5 test, I'll make not put it

### DIFF
--- a/lib/generators/statesman/active_record_transition_generator.rb
+++ b/lib/generators/statesman/active_record_transition_generator.rb
@@ -23,13 +23,5 @@ module Statesman
     def migration_file_name
       "db/migrate/#{next_migration_number}_create_#{table_name}.rb"
     end
-
-    def rails_4_or_higher?
-      Rails.version.split(".").map(&:to_i).first >= 4
-    end
-
-    def rails_5_or_higher?
-      Rails.version.split(".").map(&:to_i).first >= 5
-    end
   end
 end
--- a/lib/generators/statesman/active_record_transition_generator.rb
+++ b/lib/generators/statesman/active_record_transition_generator.rb
@@ -27,5 +27,9 @@ module Statesman
     def rails_4_or_higher?
       Rails.version.split(".").map(&:to_i).first >= 4
     end
+
+    def rails_5_or_higher?
+      Rails.version.split(".").map(&:to_i).first >= 5
+    end
   end
 end
--- a/spec/generators/statesman/migration_generator_spec.rb
+++ b/spec/generators/statesman/migration_generator_spec.rb
@@ -7,13 +7,13 @@ describe Statesman::MigrationGenerator, type: :generator do
     let(:migration_name) { 'db/migrate/add_statesman_to_bacon_transitions.rb' }
   end
 
+  let(:migration_number) { '5678309' }
+
+  let(:mock_time) do
+    double('Time', utc: double('UTCTime', strftime: migration_number))
+  end
+
   describe 'the model contains the correct words' do
-    let(:migration_number) { '5678309' }
-
-    let(:mock_time) do
-      double('Time', utc: double('UTCTime', strftime: migration_number))
-    end
-
     subject do
       file(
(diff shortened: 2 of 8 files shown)


spec/support/active_record.rb
```diff
@@ -1,6 +1,12 @@
-require "support/active_record"
```
kenchan0130: I removed this because this file need not require self.

lib/generators/statesman/active_record_transition_generator.rb
```diff
@@ -27,5 +27,9 @@ def migration_file_name
     def rails_4_or_higher?
       Rails.version.split(".").map(&:to_i).first >= 4
     end
+
+    def rails_5_or_higher?
```
greysteil: You shouldn't need this function once it's in `GeneratorHelpers`. Would be nice to move `rails_4_or_higher?` into there, too.

greysteil:
I'm not a maintainer here anymore, but this looks good to me.

hmac:
This looks good. It'd be nice to have a spec in `spec/generators/statesman/migration_generator_spec.rb` that tests we're generating the right stuff for Rails 5.

lib/generators/statesman/active_record_transition_generator.rb
kenchan0130: alright.

I'll do it.

kenchan0130: see https://github.com/gocardless/statesman/pull/243/commits/cbd609a06eabc13020ce92a5a6fc14f7c11e4636

kenchan0130:
@hmac 

Thank you for comment.
I have updated test.

spec/generators/statesman/migration_generator_spec.rb
```diff
@@ -39,4 +39,24 @@
...
+      run_generator %w(Yummy::Bacon Yummy::BaconTransition)
+    end
+
+    if ::ActiveRecord.respond_to?(:gem_version) &&
+       ::ActiveRecord.gem_version >= Gem::Version.new('5.0.0')
+      it { is_expected.to contain(/ActiveRecord::Migration\[\d+\.\d+\]/) }
+    else
+      it { is_expected.not_to contain(/ActiveRecord::Migration\[\d+\.\d+\]/) }
+    end
+  end
```
hmac: So I'd prefer to mock out `ActiveRecord.gem_version` to return `5.0.0` and expect that the output matches a static fixture, rather than testing it dynamically like this.

spec/support/active_record.rb
```diff
@@ -1,6 +1,12 @@
...
 require "json"
 
+MIGRATION_CLASS = if Rails.version.split(".").map(&:to_i).first >= 5
+                    migration_version = ActiveRecord::Migration.current_version
+                    ActiveRecord::Migration[migration_version]
+                  else
+                    ActiveRecord::Migration
+                  end
+
```
hmac: Similarly here - I'd rather either:
- update all the specs to use AR 5.0 syntax and have a single explicit test for AR < 5.0
- keep all the specs on AR 4.0 < 5.0 syntax and have an explicit test for AR 5.0

kenchan0130: I think using "update all the specs to use AR 5.0 syntax and have a single explicit test for AR < 5.0" because we can't use syntax in AR version < 5.0 and have to prepare a lot of conditional branches.

spec/generators/statesman/migration_generator_spec.rb
kenchan0130: Okey, I'll try my best!

ojab:
Any update on this?

stevebissett:
Is there anything blocking this going in? Or anything I can help with?

kenchan0130:
Superseded by #265

PR #248, opened 2017-05-18
Add foreign keys when creating a new transition table

greysteil:
Because `Statesman is ... designed to provide ... data integrity.`. :octocat:

### DIFF
--- a/lib/generators/statesman/templates/create_migration.rb.erb
+++ b/lib/generators/statesman/templates/create_migration.rb.erb
@@ -9,6 +9,7 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.timestamps null: false
     end
 
+    # Foreign keys are optional, but highly recommended
     add_foreign_key :<%= table_name %>, :<%= parent_table_name %>
 
     add_index(:<%= table_name %>,
--- a/lib/generators/statesman/templates/create_migration.rb.erb
+++ b/lib/generators/statesman/templates/create_migration.rb.erb
@@ -9,6 +9,8 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.timestamps null: false
     end
 
+    add_foreign_key :<%= table_name %>, :<%= parent_table_name %>
+
     add_index(:<%= table_name %>,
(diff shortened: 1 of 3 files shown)


lib/generators/statesman/templates/create_migration.rb.erb
```diff
@@ -9,6 +9,8 @@ class Create<%= migration_class_name %> < ActiveRecord::Migration
       t.timestamps null: false
     end
 
+    add_foreign_key :<%= table_name %>, :<%= parent_table_name %>
```
kenchan0130: I think that you should change `lib/generators/statesman/templates/update_migration.rb.erb` too.

kenchan0130:
I think that introducing foreign key constraints depends on team and project policies.
Is it more appropriate to put it in the template as a position of this gem?

lib/generators/statesman/templates/create_migration.rb.erb
greysteil: I deliberately kept it out of there, under the assumption that the end user would have already added a foreign key to the existing table they're updating if they wanted one.

greysteil:
@kenchan0130 - good point, I've updated the migration to include a comment above the foreign key addition.

lib/generators/statesman/templates/create_migration.rb.erb
kenchan0130: > under the assumption that the end user would have already added a foreign key to the existing table they're updating if they wanted one.

I see, but it depends on the situation.

kenchan0130:
@greysteil 

Your comment approach is working out a compromise.
look to me good.

Sinjo:
Agree with having them by default in the template. If people don't like foreign keys they can always delete them from the generated migration.

Sinjo:
An afterthought: I hadn't thought about `add_foreign_key` with SQLite. It turns out it's just a no-op there.

I'm not sure if we care about SQLite support (we don't test it). I'll have a chat with another maintainer or two and either add SQLite to the build or add an explicit notice in `README.md` that we only support Postgres and MySQL.

Issue #205, opened 2016-03-10
Transitions don't require a new transaction


MSch:
https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L73

If `ActiveRecord::Base.transaction` is called while already inside a transaction it does not create a "nested" transaction (actually a savepoint) c.f. [Nested Transactions](http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) in the Rails documentation

This means the following code will lead to an incorrect state in the DB:

``` ruby
MyModel.transaction do
  begin do
    my_model.transition_to(:something_that_fails)
  rescue Error
    # handle it somehow, expecting the transition to have been rolled back
    # but because we were already in a tx nothing happened, the db has been modified
    # and all guarantees are now void
  end
end
```

I actually ran into the opposite issue with aasm a while back, so the ideal solution would be to make the `requires_new` setting on the AR transaction configurable, c.f. https://github.com/aasm/aasm/issues/107

hmac:
@MSch are you able to provide steps to reproduce or (even better) a failing test case for this? What you've described makes sense in the abstract (you could get into a state where you think a transition has rolled back but it hasn't) but I'm struggling to make that happen in reality.

hmac:
I think it makes sense to have `requires_new: true` on all transactions - Rails' fake nested transactions are pretty much never what you actually want.

PR #249, opened 2017-05-26
Enforce real nested transactions

hmac:
`requires_new: true` forces Rails to create real nested transactions in
databases that support it. This keeps proper transaction semantics when
transitioning a state machine inside a larger transaction.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -72,7 +72,7 @@ module Statesman
 
         transition = transitions_for_parent.build(transition_attributes)
 
-        ::ActiveRecord::Base.transaction do
+        ::ActiveRecord::Base.transaction(requires_new: true) do
           @observer.execute(:before, from, to, transition)
           unset_old_most_recent
           transition.save!
--- a/lib/tasks/statesman.rake
+++ b/lib/tasks/statesman.rake
@@ -19,7 +19,7 @@ namespace :statesman do
     batch_size = 500
 
     parent_class.find_in_batches(batch_size: batch_size) do |models|
-      ActiveRecord::Base.transaction do
+      ActiveRecord::Base.transaction(requires_new: true) do
         if Statesman::Adapters::ActiveRecord.database_supports_partial_indexes?
           # Set all transitions' most_recent to FALSE
           transition_class.where(parent_fk => models.map(&:id)).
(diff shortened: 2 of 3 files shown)


hmac:
This is probably a breaking change, since some users may be relying on this behaviour at the moment.

hmac:
Fixes #205

timrogers:
@hmac What should be our next steps in getting this merged? Feels pretty sensible conceptually, it's really helped us at GoCardless, and it has been noticed by others (in #205, that is).

hmac:
I reckon we should merge this in as part of a new major version. It's worth bundling up any other breaking changes we want to make at the same time.

adaam2:
+1 for this. Just ran into a situation where I am running a transition in a parent transaction block. If other users of this library are depending on the `requires_new: false` behaviour, then could we at least make this configurable, as this seems to be quite a common use case to be able to rollback a transition within the context of a wider transaction :(

lib/statesman/config.rb
```diff
@@ -12,5 +12,9 @@ def initialize(block = nil)
     def storage_adapter(adapter_class)
       @adapter_class = adapter_class
     end
+
+    def requires_new_transaction(requires_new)
```
hmac: Might be clearer to call this something like `use_real_transactions`? Or invert it to `use_virtual_transactions`? The “requires new” terminology is very AR specific and probably doesn’t mean much to people who use other ORMs.

lib/statesman.rb
```diff
@@ -26,9 +26,14 @@ module Adapters
...
+    @requires_new = config.requires_new
   end
 
   def self.storage_adapter
     @storage_adapter || Adapters::Memory
   end
+
+  def self.requires_new
+    @requires_new || true
+  end
```
hmac: Would be great to have some documentation on what this config setting does and why!

lib/statesman/config.rb
danwakefield: True, I assumed since the option is only useful if you are using the AR adapter it would make sense.

danwakefield:
I suppose we could leave it unconfigured, Users could pin to the last version while we add the toggle if we really need to.

It's a small consideration but we do get to avoid some method calls and lookups on each transition if its a hardcoded default.

lib/statesman.rb
adaam2: Would it not be more sensible for it to fallback to `false`?

PR #250, opened 2017-05-26
Start list of notable third-party extensions

Sinjo:
Let's do this and close #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,10 @@ describe "some callback" do
 end
 ```
 
+# Third-party extensions
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -412,6 +412,10 @@ describe "some callback" do
 end
 ```
 
+# Third-party extensions
+
+[statesman-sequel](https://github.com/badosu/statesman-sequel) - An adapter to make Statesman work with the [Sequel gem](https://github.com/jeremyevans/sequel)
```
hmac: Thoughts on this vs `to make Statesman work with Sequel`?

PR #252, opened 2017-05-26
Drop support for old Ruby and Rails versions

hmac:


.travis.yml
```diff
@@ -29,21 +28,13 @@ script:
...
 
 env:
-  - "RAILS_VERSION=3.2.21"
-  - "RAILS_VERSION=4.0.13"
-  - "RAILS_VERSION=4.1.14"
   - "RAILS_VERSION=4.2.8"
   - "RAILS_VERSION=4.2.8 DATABASE_URL=mysql2://root@localhost/statesman_test"
   - "RAILS_VERSION=4.2.8 DATABASE_URL=postgres://postgres@localhost/statesman_test"
-  - "RAILS_VERSION=5.0.0 EXCLUDE_MONGOID=true"
+  - "RAILS_VERSION=5.0.0 EXCLUDE_MONGOID=true DATABASE_URL=mysql2://root@localhost/statesman_test"
```
Sinjo: I think we're adding 5.0.0 and 5.1.0 in other PRs. Shall we limit this one to the removal of the old versions?

hmac: We're already testing against 5.0, but yeah happy to do 5.1 in a separate PR

Sinjo: 👍

greysteil:
FYI, if you do this, you should probably set the `required_ruby_version` in the gemspec. http://guides.rubygems.org/specification-reference/#required_ruby_version

PR #253, opened 2017-05-26
Add compatibility policy for Ruby and Rails

Sinjo:
First draft. Let's get this round the team before we merge it.

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -1,0 +1,17 @@
+# Compatibility
+
+Our goal as Statesman maintainers is for the library to be compatible with all supported versions of Ruby and Rails.
+
+Specifically, any CRuby/MRI version that has not received an End of Life notice ([e.g. this notice for Ruby 2.1](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/)) is supported. Similarly, any version of Rails listed as currently supported on [this page](http://guides.rubyonrails.org/maintenance_policy.html) is one we aim to support in Statesman.
+
+To that end, [our build matrix](.travis.yml) includes all these versions.
+
+Any time Statesman doesn't work on a supported combination of Ruby and Rails, it's a bug, and can be reported [here](https://github.com/gocardless/statesman/issues).
+
+# Deprecation
+
+Whenever a version of Ruby or Rails falls out of support, we will mirror that change in Statesman by updating the build matrix and releasing a new major version.
+
+At that point, we will close any issues that only affect the unsupported version, and may choose to remove any workarounds from the code that are only necessary for the unsupported version.
+
+We will then bump the major version of Statesman, to indicate the break in compatibility. Even if the new version of Statesman happens to work on the unsupported version of Ruby or Rails, we consider compatibility to be broken at this point.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 ![Statesman](http://f.cl.ly/items/410n2A0S3l1W0i3i0o2K/statesman.png)
 
-A statesmanlike state machine library for Ruby 2.0.0 and up.
+A statesmanlike state machine library.
+
+For our policy on compatibility with Ruby and Rails versions, see [COMPATIBILITY.md](docs/COMPATIBILITY.md).
 
 [![Gem Version](https://badge.fury.io/rb/statesman.svg)](http://badge.fury.io/rb/statesman)
 [![Build Status](https://travis-ci.org/gocardless/statesman.svg?branch=master)](https://travis-ci.org/gocardless/statesman)


docs/COMPATIBILITY.md
```diff
@@ -0,0 +1,17 @@
+# Compatibility
+
+Our goal as Statesman maintainers is for the library to be compatible with all supported versions of Ruby and Rails.
+
+Specifically, any CRuby version that has not received an End of Life notice ([e.g. this notice for Ruby 2.1](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/)) is supported. Similarly, any version of Rails listed as currently supported on [this page](http://guides.rubyonrails.org/maintenance_policy.html) is one we aim to support in Statesman.
```
hmac: Thoughts on CRuby vs MRI?

Sinjo: The community is moving towards CRuby (anecdotally). We could say "CRuby/MRI" if we want to be really clear.

nickcampbell18: I like that latter. I still call it MRI.

Sinjo: Changing it now.

docs/COMPATIBILITY.md
```diff
@@ -0,0 +1,17 @@
...
+
+Any time Statesman doesn't work on a supported combination of Ruby and Rails, it's a bug, and can be reported [here](https://github.com/gocardless/statesman/issues).
+
+# Deprecation
+
+Whenever a version of Ruby or Rails falls out of support, we will mirror that change in Statesman by updating the build matrix and releasing a new major version.
+
+At that point, we will close any issues that only affect the unsupported version, and may choose to remove any workarounds from the code that are only necessary for the unsupported version.
+
+We will then bump the major version of Statesman, to indicate the break in compatibility. Even if the new version of Statesman happens to work on the unsupported version of Ruby or Rails, we consider compatibility to be broken at this point.
```
timrogers: Does bumping the major version make sense here? You can imagine a situation in which there are no actual changes to functionality or compatibility, but we just release an "empty" version.

Is this necessarily better than (a) dropping unsupported versions from our CI and (b) making clear in READMEs etc. what versions are supported?

Sinjo: > You can imagine a situation in which there are no actual changes to functionality or compatibility, but we just release an "empty" version.

So this is exactly what we're thinking of doing.

Our reasoning is that once you take those versions out of CI, you have no idea if you work with them any more. We think that a bump of the major version is the indicator most likely to get users to take stock and check the changelog.

The other options we discussed - and they're still on the table - were:

1) Publish the policy, roll Ruby/Rails out of support without releasing any kind of new version
2) Publish the policy, allow breaking compatibility in minor releases (not patch)

The reason we dismissed those was that we figured people don't go and read the compatibility policies of every library they use, and were unlikely to realise that we'd dropped support. The one thing they'd be most likely to see is the bump in the major version, which would prompt them to go investigate why it was bumped.

Honestly we're a little torn, but this seemed like the option with the best outcome for everyone. It definitely feels weird to rattle through major releases like this though, but I can't back up that feeling with an actual reason for it being bad.

I'm going to see if some friends with experience maintaining open source projects have any thoughts. Don't be surprised if some new people show up in this thread!

timrogers: Sounds super sensible to me! 

I know you guys know that what you're proposing will lead sometimes to "empty" versions, but as you say, that feels rather weird, despite me not being able to offer any real reason why it's bad per se. 

I kind of wonder whether the major version bump will have the hopes-for signalling function, as I suspect there are two kinds of people: people who stay up to date (and so will be on modern Ruby and Rails versions anyway) and people who don't (and won't!). I'm not sure a major bump therefore will jog people into action necessarily.

I definitely buy asking around.

hmac: I guess I see it more as being a good citizen wrt semver - for someone on an old version of Rails, dropping support for that version would be a breaking change (since there's now no guarantee that statesman works with their app). 

I like the security of knowing that you can (almost always) safely upgrade minor and patch versions of your dependencies without worrying about introducing incompatibilities into your app.

Sinjo: Thoughts from a friend who does a fair bit of open source:

```
18:40:07 < Sinjo> Mmm, it's hard to figure out what's "right". Like...we will drop support for those Ruby/Rails versions as they roll out of security support. There's no question around that.
18:40:34 < Sinjo> Does that mean we've broken compat and should bump major?
18:52:28 < friend> Sinjo: My gut says no, as "compat" here generally means for people calling *into* your library/app.
```

jacobpgn: This feels patch/minor to me, even though I'm usually keen to bump versions! We are not introducing any breaking changes to the API in this situation.

If dependencies resolve then you're in a good place and everything should be expected to continue working as before. If dependencies don't resolve and you first need to - for example - bump Rails, _that's_ where your breaking changes come from.

Sinjo: > If dependencies resolve then you're in a good place and everything should be expected to continue working as before.

That's a big part of the issue here. We allow people to supply adapters (such as the Sequel one), so we don't have a hard dependency on Rails or ActiveRecord. This is the only way that we have to advertise this breaking change.

> This feels patch/minor to me

For me, if we're not doing major, then there's no point to a special release when we stop supporting something. Might as well just drop it out the build matrix and crack on.

jacobpgn: That's legit.

Interesting discussions elsewhere:
https://github.com/mojombo/semver/issues/148
https://github.com/quartznet/quartznet/issues/229
https://github.com/guard/listen/issues/374

From these sources and others (looked at a bunch of issues raised by confused users across other projects) I'm getting a vibe of:
The **logical** thing to do is to bump minor/patch.
The **expected** thing to do is bump major.

If users expect that this should be a major bump, then I'm quite tempted by that approach...

Sinjo: Just so this isn't left hanging - I'm still thinking about this. I've had a skim of those threads, but I want to have spend another 30 minutes reading/thinking before making a call.

Sinjo:
Right, I've read mojombo/semver#148 and I'm convinced this is the right path.

The strongest reason for me is that we don't have a hard dependency on ActiveRecord due to the pluggable adapter system, so the major bump is the only signal of a Ruby/Rails version falling out of support.

We can always revisit this if we find it causes chaos.

Issue #254, opened 2017-06-26
Make it possible to determine which guard failed on a transition


timrogers:
## Problem

Using `Machine.guard_transition` in your state machine, you can create a "guard" which determines whether a specified transition is permitted, specifying a block to be called which returns a boolean representing whether the request transition is allowed. 

But if you have multiple guards and a transition is prevented, you can't tell which guard (and thus what logic) blocked the transition.

For example, I might have a piece of code like the following, specifying a few different guards determining whether I can transition a subscription to an "active" state:

```ruby
guard_transition(from: :cancelled, to: :active) do |subscription|
  subscription.product.reactivation_permitted?
end

guard_transition(to: :active) do |subscription|
  subscription.user.payment_method.present?
end
```

Let's say for example, that I provide an API endpoint for activating a subscription. In that endpoint's code, I want to return a helpful error explaining why I couldn't transition it to `:active` if the transition fails.

At present, you can't really do this in a sensible way. Either you'll call `Machine#transition_to` which will return `false`, or `Machine#transition_to!` which will raise a `GuardFailedError`. 

At the moment, probably the best solution to this would be to duplicate the guard logic elsewhere, having some code like this:

```ruby
return reactivation_not_permitted_error unless subscription.product.reactivation_permitted?
return no_payment_method_error unless subscription.user.payment_method.present?

subscription.transition_to!(:active)
render @subscription
```

Potentially, the logic which you end up duplicating could be fairly complex, involving separate classes or multiple chained conditions.

## Potential solutions

It'd be really nice if Statesman offered a native way to solve this problem, providing some kind of context as to what failed, rather than just a blanket, generic failure.

Here are two proposals for doing this for discussion - feedback and alternatives are welcomed:

### 1. Giving guards a name

Guards could be defined with a name:

```ruby
guard_transition(to: :active, name: :payment_method_required) do |subscription|
  subscription.user.payment_method.present?
end

begin
  subscription.transition_to!(:active)
rescue Statesman::GuardFailedError => e
  return reactivation_not_permitted_error if e.name == :reactivation_must_be_permitted
  return no_payment_method_error if e.name == :payment_method_required
end

render @subscription
```

On failure, `#transition_to!` raise a `GuardFailedError` with the `:name` attached as metadata. `#transition_to` could also potentially return this too, perhaps by returning the symbol name.

### Providing a exception to guards to be raised

When defining a guard, a subclass of `GuardFailedError` could be specified to be raised if the guard fails:

```ruby
[...]

PR #262, opened 2017-08-21
Enable determining which guard failed causing a GuardFailedError

timrogers:
Guards allow you to prevent a transition from occuring under certain conditions defined in your state machine.

When calling `Machine#transition_to!`, a `Statesman::GuardFailedError` is raised when a guard fails (i.e. returns false). As noted in #254, it is not currently possible to tell *which* Guard failed.

If you wanted to be able to different things depending on why you couldn't transition (e.g. returning a specific error response in an API), this means that you'll probably end up duplicating code, and at the very least, performing your guard operation twice, which could be expensive:

```ruby
return reactivation_not_permitted_error unless subscription.product.reactivation_permitted?
return no_payment_method_error unless subscription.user.payment_method.present?

subscription.transition_to!(:active)
render @subscription
```

This exposes a `guard_name` on `Statesman::GuardFailedError` which can be inspected and used to take relevant actions without performing the guarding operation twice.

```ruby
begin
  subscription.transition_to!(:active)
rescue Statesman::GuardFailedError => e
  return reactivation_not_permitted_error if e.guard_name == :reactivation_must_be_permitted
  return no_payment_method_error if e.guard_name == :payment_method_required
end

render @subscription
```

### DIFF
--- a/lib/statesman/guard.rb
+++ b/lib/statesman/guard.rb
@@ -3,10 +3,17 @@ require_relative "exceptions"
 
 module Statesman
   class Guard < Callback
+    attr_reader :name
+
+    def initialize(options = { from: nil, to: nil, callback: nil })
+      @name = options[:name]
+      super
+    end
+
     def call(*args)
       unless super(*args)
-        raise GuardFailedError,
-              "Guard on transition from: '#{from}' to '#{to}' returned false"
+        raise GuardFailedError.new("Guard on transition from: '#{from}' to " \
+                                   "'#{to}' returned false", guard_name: name)
       end
     end
   end
--- a/spec/statesman/machine_spec.rb
+++ b/spec/statesman/machine_spec.rb
@@ -615,8 +615,12 @@ describe Statesman::Machine do
 
     context "when it is unsuccesful" do
       before do
+        guard_failed_error = Statesman::GuardFailedError.new(
+          "Guard on transition from: 'active' to 'inactive' returned false"
+        )
+
         allow(instance).to receive(:transition_to!).
-          and_raise(Statesman::GuardFailedError)
+          and_raise(guard_failed_error)
       end
       it { is_expected.to be_falsey }
     end
--- a/lib/statesman/exceptions.rb
+++ b/lib/statesman/exceptions.rb
@@ -2,10 +2,18 @@ module Statesman
   class InvalidStateError < StandardError; end
   class InvalidTransitionError < StandardError; end
   class InvalidCallbackError < StandardError; end
-  class GuardFailedError < StandardError; end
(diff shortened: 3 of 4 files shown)


timrogers:
@hmac Can you give this a first review?

lib/statesman/guard.rb
```diff
@@ -3,10 +3,17 @@
 
 module Statesman
   class Guard < Callback
+    attr_reader :name
+
+    def initialize(options = { from: nil, to: nil, callback: nil })
```
timrogers: Is there a nicer way to do this which doesn't make the `#initialize` of `Guard` as tied to that of `Callback?`

spec/statesman/machine_spec.rb
```diff
@@ -615,8 +615,12 @@ def after_initialize; end
 
     context "when it is unsuccesful" do
       before do
+        guard_failed_error = Statesman::GuardFailedError.new(
+          "Guard on transition from: 'active' to 'inactive' returned false"
+        )
+
         allow(instance).to receive(:transition_to!).
-          and_raise(Statesman::GuardFailedError)
```
timrogers: This seems to misbehave once you add an additional parameter to the error.

lib/statesman/guard.rb
hmac: Can't you just do
```ruby
class Guard
  def initialize(options)
    @name = options.delete(:name)
    super(options)
  end
end
```

spec/statesman/machine_spec.rb
hmac: This sounds like it'll break a lot of code then...

lib/statesman/guard.rb
timrogers: Yes, could do, although you use the defaults. But the defaults are pointless, since they're `nil`. I'll change this.

spec/statesman/machine_spec.rb
timrogers: Going to experiment with this today - it seems pretty weird that it breaks it. Wondering if not using a keyword argument might help.

nickcampbell18:
👋 HI TIM. What shall we do with this PR?

dmitry:
@danwakefield what was the reason it was closed? Seems like solves an issue #254

nickcampbell18:
@dmitry I think we probably closed this due to inactivity/lack of sponsorship. Would you be interested in reviving the branch and bringing it up to date? I'd happily review & merge it if so!

dmitry:
@nickcampbell18 going to check if it's the only way to detect which guard was used. Straight forward solution might be to use AA `record.errors.add(:base, :reason)` for such case, but I don't know yet about the cons of this method.

PR #267, opened 2017-08-31
Add information about guard_transition operation.

danwakefield:
Make the behaviour of the Bool returned during a guard_transition
more obvious

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,12 +205,23 @@ an array of states (`.transition(from: :some_state, to: [:another_state, :some_o
 #### `Machine.guard_transition`
 ```ruby
 Machine.guard_transition(from: :some_state, to: :another_state) do |object|
-  object.some_boolean?
+  next false if object.broken?
+  next true if object.skip_check_thing?
+
+  object.check_thing?
 end
 ```
 Define a guard. `to` and `from` parameters are optional, a nil parameter means
 guard all transitions. The passed block should evaluate to a boolean and must
 be idempotent as it could be called many times.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -212,6 +212,10 @@ Define a guard. `to` and `from` parameters are optional, a nil parameter means
 guard all transitions. The passed block should evaluate to a boolean and must
 be idempotent as it could be called many times.
 
+Returning `false` indicates failure within the Guard and raises a `GuardFailedError`.
```
timrogers: Does it not only raise the error if you use `Machine#transition_to!`?

README.md
```diff
@@ -205,13 +205,24 @@ an array of states (`.transition(from: :some_state, to: [:another_state, :some_o
...
+
+  object.check_thing?
 end
 ```
 Define a guard. `to` and `from` parameters are optional, a nil parameter means
 guard all transitions. The passed block should evaluate to a boolean and must
 be idempotent as it could be called many times.
 
+If the block evaluates as `false` the Guard fails. This will raise a
+`GuardTransitionError` when using `transition!` or return `false` when using
```
timrogers: The method which raises is called `#transition_to!` (sorry!)

README.md
```diff
@@ -205,13 +205,24 @@ an array of states (`.transition(from: :some_state, to: [:another_state, :some_o
...
+`transition`.
```
timrogers: Probably safest to say that the other methods all don't raise - there are tonnes of options!

PR #268, opened 2017-08-31
	Build Statesman using CircleCI 2.0

timrogers:
This switches to using CircleCI 2.0 to build Statesman.

This involves using a fairly complex Workflows setup with lots of different jobs for each Ruby version, Rails version and database adapter. It's a little more complicated to set up than Travis, since there's no matrix support, but it's faster and more aligned with our tooling at GoCardless for everything else. We end up with a mammoth 21 job workflow 🙈🐘

We were previously missing the two database adapters for Rails 5.0 and 5.1, so this also adds those.

It's worth noting that it doesn't seem to be possible to achieve better reuse (i.e. a more DRY approach) with the different images in the config (e.g. the standard setup for MySQL, Postgres).
See https://stackoverflow.com/questions/4948933/is-there-a-way-to-alias-anchor-an-array-in-yaml for details. It doesn't appar that you can have a key that contains multiple merge keys as an array.


timrogers:
@hmac All yours!

.circleci/config.yml
```diff
@@ -0,0 +1,429 @@
...
+      - image: circleci/ruby:2.4.1-node
+        environment:
+          - RAILS_VERSION=5.0.5
+          - DATABASE_URL=postgres://postgres@localhost/statesman_test
+          - EXCLUDE_MONGOID=true
+      - image: circleci/postgres:9.6
+        environment:
+          - POSTGRES_USER=postgres
+          - POSTGRES_DB=statesman_test
+      - image: circleci/mysql:5.7.18
```
hmac: could we drop the mysql container if we're running against postgres here?

.circleci/config.yml
```diff
@@ -0,0 +1,429 @@
...
+          - MYSQL_DATABASE=statesman_test
+          - MYSQL_ROOT_HOST: "%"
+    steps: *steps
+  build-ruby241-rails-429-mysql:
+    docker:
+      - image: circleci/ruby:2.4.1-node
+        environment:
+          - RAILS_VERSION=4.2.9
+          - DATABASE_URL=mysql2://root@127.0.0.1/statesman_test
+      - image: circleci/postgres:9.6
```
hmac: same here, can we drop postgres if we're just testing against mysql?

.circleci/config.yml
```diff
@@ -0,0 +1,429 @@
...
+          - MYSQL_USER=root
+          - MYSQL_PASSWORD=
+          - MYSQL_DATABASE=statesman_test
+    steps: *steps
+
+workflows:
+  version: 2
+  tests:
+    jobs:
+      - build-ruby241-rails-429
```
hmac: How come we have a standalone build for rails `4.2.9` but no equiavalent for `5.0.5` or `5.1.3`?

timrogers: We came to the conclusion that the purpose of this build is to test the Mongo adapter. The rest of the tests run in every case, so there is a fair bit of duplication.

An interesting future refactor would be to split up the tests a bit more logically so you can have a standalone set of tests and then tests for each adapter. This would be quite a bit of work, but would make more sense.

I'm converted the standalone build to explicitly be a Mongo build, and have added a Mongo build for Rails 5.0.x and 5.1.x, which were conspicuously absent.

timrogers:
@hmac This should be ready now! I've taken away extraneous database containers where they're not needed, renamed the apparently-standalone test to explicitly be a Mongo test, added Mongo tests for Rails 5.0.x and 5.1.x and started using `dockerize` to reduce flakiness by waiting for the databases to be up.

PR #276, opened 2017-10-30
allowed_transitions? accept metadata to pass to guards

vvondra:
Internally, allowed_transitions? uses can_transition_to? to call all guards on a transition to check it's feasability.

While you can pass metadata into can_transition_to? for the guards to use, you don't have the possibility on allowed_transitions?

This allows you to list the allowed transitions given a metadata context (we use metadata in our transitions heavily to add context of the change)

There are no breaking changes.

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -189,9 +189,9 @@ module Statesman
       states.flatten.any? { |state| current_state == state.to_s }
     end
 
-    def allowed_transitions
+    def allowed_transitions(metadata = {})
       successors_for(current_state).select do |state|
-        can_transition_to?(state)
+        can_transition_to?(state, metadata)
       end
     end
 
--- a/spec/statesman/machine_spec.rb
+++ b/spec/statesman/machine_spec.rb
@@ -441,7 +441,8 @@ describe Statesman::Machine do
     end
 
     let(:instance) { machine.new(my_model) }
-    subject { instance.allowed_transitions }
+    let(:metadata) { { some: :metadata } }
+    subject { instance.allowed_transitions(metadata) }
 
     context "with multiple possible states" do
       it { is_expected.to eq(%w[y z]) }
@@ -450,6 +451,16 @@ describe Statesman::Machine do
     context "with one possible state" do
       before { instance.transition_to!(:y) }
       it { is_expected.to eq(['z']) }
+
+      context "guarded using metadata" do
+        before { machine.guard_transition(to: :z)  { |_, _, metadata | metadata[:some] == :metadata } }
+        it { is_expected.to eq(['z']) }
+      end
+
+      context "excluded by guard using metadata" do
+        before { machine.guard_transition(to: :z)  { |_, _, metadata | metadata[:some] != :metadata } }
+        it { is_expected.to eq([]) }
+      end
     end
 
     context "with no possible transitions" do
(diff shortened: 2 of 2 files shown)


vitorreis:
+1 👍  on this change

timrogers:
@hmac I'd be interested to hear your thoughts on this -- seems eminently sensible to me though!

@vvondra Would you be open to implementing this yourself?

vvondra:
@timrogers sure, I'd just add a test case on top, anything else to add?

hmac:
This makes a whole lotta sense to me. A test would be great 🙏

barisbalic:
I can't tell if I'm reading this correctly, but the intended purpose of metadata was to record things against the transition that are interesting.  This sounds like you are suggesting the metadata is being used in determining if a transition is valid, which is something I'm sure we used to advise against.

Admittedly it's then strange that it's also possible to pass metadata to the `can_transition_to?` method, but that probably shouldn't be possible either.

timrogers:
I can see your point @barisbalic, but given that we already support this on `Machine#can_transition_to?` and it feels like a legitimate use case, I think I'd be pretty keen to at least be consistent and support it across the query methods. Chatted to @hmac some more and we both agreed on that.

@vvondra Sounds great to me! Just tag me in your PR.

vvondra:
@barisbalic I would be surprised if we were the only ones to use this. One example is that we have states which are temporary, the metadata holds the time until the state can be held (for example temporary break of an employee). We would validate the metadata on transition to break and also use it on transitioning back to working when the timestamp expires. You can get pretty creative.

Either way it's now part of the public API (albeit the second and third parameters passed two `before_transition` and `guard_transition` are not very-well documented), I'd be happy if it was consistent.

vvondra:
Tests added and passing.

@timrogers this is already the PR :)

barisbalic:
@vvondra that sounds a bit sticky, the original guidance I would have given (and you probably considered) is that some external entity should schedule the state changes.

@timrogers @hmac I understand the desire for consistency, but I would say that is better achieved by remove the capacity to provide it to `can_transition_to?`.  So I agree consistency is important, but I would suggest the right thing to do is actually to remove the possibility altogether.

The situation @vvondra finds himself in is exactly the kind of thing we try to discourage, the state machine is supposed to make state transitions and enforcement simple and easy to understand at a glance.  This metadata business is allowing people to make it very complicated, and is something we've actively tried to avoid as a principle of Statesmans design.

I strongly feel like this is a step in the wrong direction...

timrogers:
Thanks so much for your PR @vvondra, and for your thoughts @barisbalic - we'll chat a bit more internally about this feature, and will try to come back soon.

barisbalic:
@timrogers no problem, happy to jump on some form of chat app if you wanted to discuss further.

timrogers:
Thanks for your patience on this @vvondra!

I'm sorry that it's taken so long to be able to get a proper response back to your PR, but we wanted to be able to take the time to chat this through properly at GoCardless (with me, @hmac and @petehamilton), but people have been in and out on holidays and the like.

Whilst we agree with @barisbalic that allowing or disallowing transitions based on metadata somewhat goes against the design philosophy behind Statesman (they were *originally* there to record interesting information, not allow or disallow the transition), we have supported this way of using the library since the second ever Statesman PR (see https://github.com/gocardless/statesman/pull/2), where we started to make metadata accessible from guards.

Given that we've already done this, it's well bedded-in (indeed, we use this at GoCardless ourselves) and `Machine#can_transition_to?` supports it, it feels only right to "square the circle" and consistently support this way of working in `Machine#allowed_transitions?` rather than pretending the library isn't used in this way.

Given that, we're going to merge this PR and release the change as a minor version bump.

We would, however, like to think about deprecating this in a future major version - but only once we've tested and can advocate for an alternative implementation for this use case.

vvondra:
Sounds good to me, thanks for the transparency!

barisbalic:
:+1: good shout gentlemen.

PR #278, opened 2017-11-15
Enable inheritance between state machine classes

Sannim1:
In order to use a state machine for tracking the various states of your application object, Statesman requires that you implement state machine classes. These classes are expected to `include` the `Statesman::Machine` module and make use of its class methods to define a  set of valid states, transitions and callbacks.

Depending on the application domain, it might be required that you implement different state machine classes for the same object based on its type. As these classes are bound to share some states, transitions and/or callbacks, this inherently results in duplicating the code which defines these states, transitions and/or callbacks in each of the state machine classes.

Since the calls to define the states, transitions and callbacks of a state machine class are executed when Ruby loads the file containing the class, it is not possible to use Ruby's inbuilt inheritance mechanism to share these definitions between classes.

Here, we introduce a Module that facilitates the sharing of such definitions between state machine classes. This can be included in any state machine class and its functions can be invoked at load time to inherit the definitions of states, transitions and callbacks from another state machine class.

### DIFF
--- a/lib/statesman.rb
+++ b/lib/statesman.rb
@@ -1,10 +1,11 @@
 module Statesman
-  autoload :Config,     'statesman/config'
-  autoload :Machine,    'statesman/machine'
-  autoload :Callback,   'statesman/callback'
-  autoload :Guard,      'statesman/guard'
-  autoload :Utils,      'statesman/utils'
-  autoload :Version,    'statesman/version'
+  autoload :Config,               'statesman/config'
+  autoload :Machine,              'statesman/machine'
+  autoload :Callback,             'statesman/callback'
+  autoload :Guard,                'statesman/guard'
+  autoload :Utils,                'statesman/utils'
+  autoload :Version,              'statesman/version'
+  autoload :MachineInheritance,   'statesman/machine_inheritance'
   module Adapters
     autoload :Memory,       "statesman/adapters/memory"
     autoload :ActiveRecord, "statesman/adapters/active_record"
--- a/lib/statesman/machine_inheritance.rb
+++ b/lib/statesman/machine_inheritance.rb
@@ -1,0 +1,47 @@
+module Statesman
+  module MachineInheritance
+    def self.included(receiver)
+      receiver.extend ClassMethods
+    end
+
+    module ClassMethods
+      def inherit_from(state_machine)
+        inherit_initial_state_from  state_machine
+        inherit_states_from         state_machine
+        inherit_transitions_from    state_machine
+        inherit_callbacks_from      state_machine
+      end
+
+      def inherit_initial_state_from(state_machine)
+        state state_machine.initial_state, initial: true
+      end
+
+      def inherit_states_from(state_machine)
+        state_machine.states.each do |state_to_inherit|
+          next if states.include? state_to_inherit
+          state state_to_inherit
+        end
+      end
+
+      def inherit_transitions_from(state_machine)
+        state_machine.successors.each do |from, to|
+          transition from: from, to: to
+        end
+      end
+
+      def inherit_callbacks_from(state_machine)
+        state_machine.callbacks.each do |callback_type, callbacks|
+          callbacks.each do |callback|
+            add_callback(
+              callback_type: callback_type,
+              callback_class: callback.class,
+              from: callback.from,
+              to: callback.to,
+              &callback.callback
+            )
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


barisbalic:
Hi @Sannim1,

It's an interesting proposal, any chance you can provide some basic concrete examples?  I'm having a hard time visualising a scenario where relying on a `type` field rather than have distinct classes for each 'type' makes sense.

Let's say for some reason you have a `Product` table, and there are two types, `subscription` and `purchase`.  You can just have a state machine per 'type'.  So you could have...

```ruby
class SubscriptionProductStateMachine
  include Statesman::Machine
  ...
end

class PurchaseProductStateMachine
  include Statesman::Machine
  ...
end

class Product < ActiveRecord::Base
  include Statesman::Adapters::ActiveRecordQueries
  ...
  def state_machine
    @state_machine ||= (type == 'purchase' ? PurchaseProductStateMachine.new(...) : SubscriptionProductStateMachine.new(...)
  end
end
```

With regards to sharing states, transitions and callbacks, I'm definitely not keen.  Part of the design philosophy for Statesman is to make state modelling as simple to understand as possible.  It is by intention that you can eyeball a state machine and see exactly what can happen and when.

To work around this, you can do exactly what you have done, introduce a module and mix in the things you want to share.  Although rather than do it like this, I would think it better to explicitly create a module for each collection of things (grouped by your concept of `type`) you want to share and then explicitly mix them in.

There's nothing wrong with your code specifically, but it seems like something people can already do and something we'd probably advise them not to do.

Sannim1:
Hi @barisbalic

Thanks for taking a look at this. 

Your example is a perfect illustration of what I meant by choosing state_machine classes according to the `type` of the domain object and this is how it is currently implemented in my use case (with just 3 types).

With a few possible types, this is all well and good. The problem arises when the domain object can be one of a large number of types(say 15). Since, after all, we are dealing with the same object (`Product` from your example), we end up having different state machine classes which all start from a `created` state and can transition into a `pending_payment` state like so:

```ruby
class SubscriptionProductStateMachine
  include Statesman::Machine

  state :created, initial: true
  state :pending_payment
  state :pending_delivery
  state :delivered

  transition from: :created, to: :pending_payment
  transition from: :pending_payment, to: :pending_delivery
  transition from: :pending_delivery, to: delivered

  ...
  # some custom callbacks for subscription products
  ...

  after_transition(to: :pending_delivery) { send_email }
end

class PurchaseProductStateMachine
  include Statesman::Machine

  state :created, initial: true
  state :pending_payment
  state :pending_delivery
  state :delivered

  transition from: :created, to: :pending_payment
  transition from: :pending_payment, to: :pending_delivery
  transition from: :pending_delivery, to: delivered

  ...
  # some custom callbacks for purchase products
  ...

  after_transition(to: :pending_delivery) { send_email }
end
```

Creating a generic product state machine containing the common states, transitions and callbacks reduces duplication and gives a clear view of the differences between the state machines. So, the above becomes:

```ruby
class GenericProductStateMachine
  include Statesman::Machine

  state :created, initial: true
  state :pending_payment
  state :pending_delivery
  state :delivered

  transition from: :created, to: :pending_payment
  transition from: :pending_payment, to: :pending_delivery
  transition from: :pending_delivery, to: delivered

  after_transition(to: :pending_delivery) { send_email }
end

class PurchaseProductStateMachine
  include Statesman::Machine
  include Statesman::MachineInheritance

  inherit_from GenericProductStateMachine

  # custom callbacks for purchase products
end

class SubscriptionProductStateMachine
  include Statesman::Machine
  include Statesman::MachineInheritance

  inherit_from GenericProductStateMachine

  # custom callbacks for subscription products
end
```

And makes it easier to add new state machine classes in a case when there can be many new `Product` types.

jc-langevin:
for those who want to do the same, It is totally possible to do the same thing without using this pull request
`module CommonProductStateMachine
    def self.included(base)
      base.state :created, initial: true
      base.state :pending_payment
      base.state :pending_delivery
      base.state :delivered

      base.transition from: :created, to: :pending_payment
      base.transition from: :pending_payment, to: :pending_delivery
      base.transition from: :pending_delivery, to: :delivered

      base.after_transition(to: :pending_delivery) { send_email }
  end
end

class PurchaseProductStateMachine
  include Statesman::Machine
  include CommonProductStateMachine

  # custom callbacks for purchase products
end

class SubscriptionProductStateMachine
  include Statesman::Machine
  include CommonProductStateMachine

  # custom callbacks for subscription products
end`

danwakefield:
Yeah we ended up using a module which we extend. e.g

```ruby
module Generic
  def define_states
    state :created, initial: true
    state :complete
  end
  
  def define_transitions
    transition from: :created, to: :complete
  end
  
 def define_callbacks
    after_transition(to: :complete) { send_email }
 end
end

def Foo
  include Statesman::Machine
  extend Generic
  
  define_states
  define_transitions
  define_callbacks

  ...
end
```

Probably should have closed this a while ago.

PR #285, opened 2017-12-15
Touch updated_at when unseting most recent flag

### LINKED ISSUE #284: Setting most_recent doesn't change updated_at
The Active Record adapter uses update_all to unset the `most_recent` boolean.
This results in the `updated_at` column not being accurate - causing us issues when using [stitch](https://www.stitchdata.com/) ELT which relies on the `updated_at` for detecting row changes in the DB.

https://github.com/gocardless/statesman/blob/c12a30309ce2a99240eb499f4f06b5c0e7e72e2b/lib/statesman/adapters/active_record.rb#L97-L101

NGMarmaduke:
Fixes #284

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -87,6 +87,12 @@ module Statesman
       end
 
       def unset_old_most_recent
+        most_recent = transitions_for_parent.where(most_recent: true)
+        updated_at = if ActiveRecord::Base.default_timezone == :utc
+                       Time.now.utc
+                     else
+                       Time.now
+                     end
         # Check whether the `most_recent` column allows null values. If it
         # doesn't, set old records to `false`, otherwise, set them to `NULL`.
         #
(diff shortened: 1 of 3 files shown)


lib/statesman/adapters/active_record.rb
```diff
@@ -87,6 +87,12 @@ def transitions_for_parent
       end
 
       def unset_old_most_recent
+        most_recent = transitions_for_parent.where(most_recent: true)
+        updated_at = if ActiveRecord::Base.default_timezone == :utc
```
danwakefield: Worth noting to other reviewers that this is how [activerecord does timestamps](https://github.com/rails/rails/blob/5-1-stable/activerecord/lib/active_record/timestamp.rb#L78). I checked incase we where going to miss some strange timezone stuff.

NGMarmaduke: Yup, it's a shame its not a public method - would be useful when using `update_all`

nickcampbell18:
Thank you for the contribution!

PR #286, opened 2017-12-16
Clean up rubocop config, and fix offense

greysteil:
Fixes current CI failure, and remove unnecessary rubocop inheritance.


nickcampbell18:
Thanks @greysteil! We've upgraded our Rubocop config in https://github.com/gocardless/statesman/pull/292 so I think we're all set here.

PR #287, opened 2017-12-16
Add readme hints on transition validation and after_destroy hooks

greysteil:
I always end up adding these to my transition classes. Perhaps useful for have them in the README?


petehamilton:
A little extra hinting sounds like a plan to me. Merging.

PR #288, opened 2017-12-16
Make generators more rubocop friendly

greysteil:
- Statesman only support Ruby 2.2+, so we can use `%i(...)` syntax for symbol arrays
- Using single quotes and double quotes inconsistently was never necessary

Won't make the generators fit everyone's rubocop setup, but will mean they have a chance of fitting some people's.


greysteil:
@petehamilton - can I get a quick review on this one, too?

petehamilton:
What's going on with the CI build 🤔 

Circle just seems to have put everything as `pending`, super weird.

greysteil:
No idea. Unlikely to be related to this PR...

PR #289, opened 2017-12-18
Fix name spacing error in ActiveRecord adapter

NGMarmaduke:
The active record adapter is currently erroring with `uninitialized constant Statesman::Adapters::ActiveRecord::Base` due to 61f0e52.

It looks like CI has stopped running for this repo - the badge is linking to a 404 page.


danwakefield:
Thanks for that. Little annoying the the PR tests don't notice failures.

PR #290, opened 2017-12-18
Update Rubocop exceptions

danwakefield:
We want to allow rspec change blocks so disabling this test
makes sense.


danwakefield:
Merging with failing tests as they are fixed in another PR (https://github.com/gocardless/statesman/pull/291

PR #291, opened 2017-12-18
Fix time dependant test.

danwakefield:
Mysql does not seem to have enough precision in timestamp to record the
difference between two updates in the tests. Force the update to occur
seconds into the future to capture the change.


dmagliola:
:+1:

PR #295, opened 2018-01-04
Fix force reloading last transition from database

jacobpgn:
The [current implementation](https://github.com/gocardless/statesman/blob/14250521d38449244e9bcb8f5b31b76e26614c39/lib/statesman/adapters/active_record.rb#L44-L60) for force reloading will use the
in-memory transitions if they're available instead of forcing a DB query
like you'd expect when passing this flag.

This means that `force_reload: true` currently forces a cache-level
reload of the most recent transition, but there is no way to force a
DB-level reload when the transitions are in memory.

In most cases the current behaviour works as expected – accessing
`model.state_machine.current_state` will issue a SELECT with LIMIT 1,
so the full set of the model's transitions will not be loaded into
memory and subsequent calls to `current_state` will issue fresh queries.
This breaks when you actually load the association, e.g. with
`model.history.inspect`. Once this has happened, the adapter's `history`
(and thus also methods such as `model.state_machine.current_state`) will
use the in-memory transitions whether `force_reload` is true or false,
and never issue fresh queries.

In practice, you often want a way to reload the transitions from the DB
to ensure that you pick up on any new transitions from other processes,
or other instances of the same model in the same process. This is what
is expected when you explicitly ask statesman to `force_reload`!

This change simply passes `force_reload` along to `history`, and checks
that it's false before going ahead and using in-memory transitions.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -41,8 +41,8 @@ module Statesman
         @last_transition = nil
       end
 
-      def history
-        if transitions_for_parent.loaded?
+      def history(force_reload: false)
+        if transitions_for_parent.loaded? && !force_reload
           # Workaround for Rails bug which causes infinite loop when sorting
           # already loaded result set. Introduced in rails/rails@b097ebe
           transitions_for_parent.to_a.sort_by(&:sort_key)
@@ -53,7 +53,7 @@ module Statesman
 
       def last(force_reload: false)
         if force_reload
-          @last_transition = history.last
+          @last_transition = history(force_reload: true).last
         else
           @last_transition ||= history.last
         end
--- a/lib/statesman/adapters/memory.rb
+++ b/lib/statesman/adapters/memory.rb
@@ -4,7 +4,6 @@ module Statesman
   module Adapters
     class Memory
       attr_reader :transition_class
-      attr_reader :history
       attr_reader :parent_model
 
       # We only accept mode as a parameter to maintain a consistent interface
@@ -30,6 +29,10 @@ module Statesman
 
       def last(*)
         @history.sort_by(&:sort_key).last
+      end
+
+      def history(*)
+        @history
       end
 
       private
--- a/lib/statesman/adapters/mongoid.rb
+++ b/lib/statesman/adapters/mongoid.rb
@@ -32,7 +32,7 @@ module Statesman
         @last_transition = nil
       end
 
-      def history
+      def history(*)
         transitions_for_parent.asc(:sort_key)
       end
 
--- a/spec/statesman/adapters/active_record_spec.rb
+++ b/spec/statesman/adapters/active_record_spec.rb
@@ -234,17 +234,35 @@ describe Statesman::Adapters::ActiveRecord, active_record: true do
           described_class.new(MyActiveRecordModelTransition, model, observer)
         end
 
-        before { alternate_adapter.create(:y, :z, []) }
-
         it "still returns the cached value" do
+          alternate_adapter.create(:y, :z, [])
+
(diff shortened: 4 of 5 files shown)


hmac:
I think this is a good idea. With `force_reload` and the AR adapter you are invariably going to want to fetch the latest transition from the database rather from the AR cache. 

I reckon this warrants a minor version bump (because of the change in behaviour) and a clear explanation in the changelog.

petehamilton:
Bonus points for a great branch name.

jacobpgn:
Thanks team!

I've squashed and updated the commit message – chatting with @petehamilton earlier we realised the description of the cause wasn't quite right. It should be correct now, so I'd love it if you could do a quick pass on the wording and see if it makes sense to you.

I'm planning to roll this and https://github.com/gocardless/statesman/pull/285 out as a minor release (https://github.com/gocardless/statesman/pull/285 is a non-breaking feature, and this PR is a bugfix). Seem legit? cc @timrogers

timrogers:
I wonder if we can find a clearer title for the commit which specifies what exactly our "fix" is - how about something like "Force reload from the DB, not the cache, in the ActiveRecord adapter" (I tried to get this down to [50 characters](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) but I couldn't do this while also making it declarative, and I'd rather have an informative message than an artificially short one).

petehamilton:
@timrogers If I can't make 50, I generally aim for `<= 72` which is where Github starts displaying things weirdly and is also less than the standard `80` character wrapping width for most people's setup.

timrogers:
Just realised my idea isn't a very good description of the substance of the change. Let me have a think.

petehamilton:
> Fix force_reload for active record models with preloaded transitions?

lib/statesman/adapters/active_record.rb
```diff
@@ -41,8 +41,8 @@ def create(from, to, metadata = {})
         @last_transition = nil
       end
 
-      def history
-        if transitions_for_parent.loaded?
+      def history(force_reload: false)
```
timrogers: Is this not slightly odd, in that it means that the interface of the ActiveRecord adapter differs from the other adapters? `#history` is a public part of the adapter interface, and the `Mongoid` and `Memory` adapters won't support this additional parameter. It might be good to expand this change to them too.

lib/statesman/adapters/active_record.rb
```diff
@@ -53,7 +53,7 @@ def history
 
       def last(force_reload: false)
```
timrogers: Do we want to make sure the other adapters don't have this problem? The memory adapter is safe:

https://github.com/gocardless/statesman/blob/b962ed219e6b05328694def8f9a1d5a504ec5c2e/lib/statesman/adapters/memory.rb#L31-L33

...but the Mongoid adapter has the same bug:

https://github.com/gocardless/statesman/blob/b962ed219e6b05328694def8f9a1d5a504ec5c2e/lib/statesman/adapters/mongoid.rb#L39-L45

lib/statesman/adapters/active_record.rb
jacobpgn: I don't think so: each of the adapters currently has a different interface! If you're explicitly calling the adapter's methods then you should know what interface you have available for that adapter.

timrogers: I'm not convinced that's quite true - e.g. `Memory#last` seems to accept any argument, presumably for compatibility with other adapters:

https://github.com/gocardless/statesman/blob/b962ed219e6b05328694def8f9a1d5a504ec5c2e/lib/statesman/adapters/memory.rb#L31

PR #306, opened 2018-01-29
Make the "updated timestamp column" for an ActiveRecord transition configurable

### LINKED ISSUE #305: Mysql2::Error: Unknown column 'transitions.updated_at' in 'field list'
I think issue #285 caused my problem, if I update vom 3.2.0 to 3.3.0.

With gem version 3.3.0 I get following error:

> ActiveRecord::StatementInvalid:
       Mysql2::Error: Unknown column 'transitions.updated_at' in 'field list': UPDATE `transitions` SET `transitions`.`most_recent` = NULL, `transitions`.`updated_at` = '2014-02-28 00:00:00.187677' WHERE `transitions`.`order_id` = 33414634 AND `order_transitions`.`most_recent` = 1

The issue is, that my table "transitions" has wilfully no attribute "updated_at", but the gem does not care about that! Please fix this.

timrogers:
In #285, we introduced a change such that, when transitioning a model to a new state, we go back to its old transitions and reset their `most_recent` using `ActiveRecord::Relation#update_all` (relying on the database to find the old transitions, bypassing ActiveRecord's in memory data), whilst also touching their `updated_at` timestamp to show that we've updated those records.

However, we can't necessarily assume that these transition models will have an `updated_at`. They might have an `updated_on` ([1][]), nothing or something else entirely.

This makes the "updated timestamp column" configurable as part of the transition, like this, defaulting to the Rails default, `:updated_at`:

```ruby
class MyActiveRecordModelTransition < ActiveRecord::Base
  include Statesman::Adapters::ActiveRecordTransition

  self.updated_timestamp_column = :updated_on

  belongs_to :my_active_record_model
end
```

We do this by defining a `class_attribute` on `Statesman::Adapters::ActiveRecordTransition`.

Fixes #305.

[1]: http://api.rubyonrails.org/classes/ActiveRecord/Timestamp.html#method-i-timestamp_attributes_for_update_in_model


timrogers:
This might be a horrific idea, but it feels like the way to fix this bug whilst not losing some nice Rails magic.

lib/statesman/adapters/active_record.rb
```diff
@@ -125,6 +121,21 @@ def transition_conflict_error?(e)
...
+                    else
+                      Time.now
+                    end
+
+        updated_timestamp_columns.
+          each_with_object({}) { |column_name, options| options[column_name] = timestamp }
+      end
+
+      def updated_timestamp_columns
+        @transition_class.send(:timestamp_attributes_for_update_in_model)
```
timrogers: It might be a terrible idea to depend on this private method. Other options include:

* Using the public `column_names` method, and assuming that *we just know* what the column names we should be interested in are (i.e. hard-coding `updated_at` and `updated_on`)
* Requiring that you have an `updated_at` column (thereby adding additional requirements over and above Rails)
* Making the updated timestamp column somehow configurable (thereby duplicating Rails' config)

hmac:
I wonder if a less dangerous approach here is to default to all of `updated_at/on`, but allow that to be overridden by a particular machine. We can then document that we'll automatically update `updated_at/on`, instead of just saying "we'll automatically update whatever AR currently decides is the right set of columns to update, and we can't guarantee what those will be".

NGMarmaduke:
Seeing as there *should* only be one most recent transition there is the potentially to change the logic from `update_all` to a simple `update` - let rails handle the timestamp logic.

The obvious side effect is all callbacks will be triggered on the last transition

timrogers:
Sadly it looks like this might have disappeared for at least some Rails versions, so this magical way is a non-starter 🔮 I reckon we should probably default to `updated_at`, since that's the Rails default which 99% of people will want, but allow it to be overridden to `nil` or `:another_column`.

What do we all think?

timrogers:
@NGMarmaduke That would be a nice way to be able to do it, but how ActiveRecord associations work, holding an object in memory, means that it won't work as expected in some cases. The issue is what we end up calling `#update` on - it might not be the freshest record from the database (i.e. since you loaded the "last transition", a new "last transition" might have been created).

Using `Relation#update_all` means we rely on the database to do the updating, rather than risk having a stale object in memory.

timrogers:
Do we think that this new setting for the updated timestamp column should be part of the state machine (e.g. passed in when you call `Machine.new`), or part of the transition (e.g. set through some kind of DSL method defined by `ActiveRecordTransition`)?

I'd be keen to go for the latter since (a) this functionality is specific to ActiveRecord (for now), and isn't a general option for other kinds of adapter/transition and (b) the columns we have is clearly related to the transition

timrogers:
@dmagliola @danwakefield @hmac Ready for review again with a new approach, where the specs actually pass!

danwakefield:
@timrogers Have to agree with daniel, I'm not sure what this is actually doing or how it should be used. Can we update the README to illustrate it?

lib/statesman/adapters/active_record_transition.rb
```diff
@@ -3,8 +3,20 @@
...
+      DEFAULT_UPDATED_TIMESTAMP_COLUMN = :updated_at
+
+      extend ActiveSupport::Concern
+
+      included do
+        serialize :metadata, JSON
+
+        class_attribute :updated_timestamp_column
+
+        def self.reset_updated_timestamp_column
```
danwakefield: I don't really understand why a user ever need / want this? If they do `updated_timestamp_column = :foo` they can just remove the line to use the default column right?

timrogers: I have this primarily to provide a way to reset the setting to its default state in tests, without depending on knowing the setting name and the correct default. If people don't like this, I could just refer to `DEFAULT_UPDATED_TIMESTAMP_COLUMN`.

timrogers:
I've now added pointers to how this works in the README, and in the transition class and migration generated by our generators.

lib/statesman/adapters/active_record_transition.rb
nickcampbell18: +1 for getting rid of this unwanted method.

spec/spec_helper.rb
```diff
@@ -11,6 +11,7 @@
 require "rspec/rails"
 require "support/active_record"
 require "rspec/its"
+require "pry"
```
nickcampbell18: Was this intentional?

timrogers: And I made a separate commit to do this (24195ad45e5653c3b92f7b84bf050f6e6477aa29) - just feels like nice thing to have in specs.

lib/statesman/adapters/active_record.rb
```diff
@@ -125,6 +120,18 @@ def transition_conflict_error?(e)
         e.message.include?(@transition_class.table_name) &&
           (e.message.include?("sort_key") || e.message.include?("most_recent"))
       end
+
+      def updated_timestamp_options
```
nickcampbell18: Rather than a method which returns a hash, and then merging the hash, why don't you just do something like:

```ruby
def with_timestamp(params = {})
  return params if @transition_class.updated_timestamp_column.nil?
end
```

This prevents a merge, which is slightly faster, and might be better style (but I'm not overly fussed).

README.md
```diff
@@ -137,6 +137,10 @@ Generate the transition model:
 $ rails g statesman:active_record_transition Order OrderTransition
 ```
 
+If you're using the ActiveRecord adapter and decide not to include the default updated
+timestamp column in your transition table, you'll need to configure the
+`updated_timestamp_column` option on the transition class.
+
```
nickcampbell18: I would give the example here of how to disable it.

timrogers:
Let's get a last review from @hmac, and then I think this will be ready to go.

README.md
```diff
@@ -137,6 +137,11 @@ Generate the transition model:
 $ rails g statesman:active_record_transition Order OrderTransition
 ```
 
+If you're using the ActiveRecord adapter and decide not to include the default updated
+timestamp column in your transition table, you'll need to configure the
+`updated_timestamp_column` option on the transition class, setting it to another column
+name (e.g. `:updated_on`) or `nil`.
+
```
hmac: > decide not to include the default `updated_at` timestamp column ?

lib/generators/statesman/templates/active_record_transition_model.rb.erb
```diff
@@ -1,6 +1,13 @@
 class <%= klass %> < <%= Statesman::Utils.rails_5_or_higher? ? 'ApplicationRecord' : 'ActiveRecord::Base' %>
   include Statesman::Adapters::ActiveRecordTransition
 
+  # If your transition table doesn't have the default updated timestamp column
```
hmac: again here?

lib/statesman/adapters/active_record.rb
```diff
@@ -1,5 +1,4 @@
 require_relative "../exceptions"
-
```
hmac: was this intentional?

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -163,6 +163,39 @@
...
+
+          after do
+            MyActiveRecordModelTransition.updated_timestamp_column =
+              MyActiveRecordModelTransition::DEFAULT_UPDATED_TIMESTAMP_COLUMN
+          end
+
+          it "just updates the most_recent" do
+            expect { Timecop.freeze(Time.now + 5.seconds) { create } }.
+              to(change { previous_transition.reload.most_recent })
+          end
```
hmac: Should we test here that we're not updating `updated_at`?

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -163,6 +163,39 @@
...
+        end
+
+        context "with no updated timestamp column set" do
+          before { MyActiveRecordModelTransition.updated_timestamp_column = nil }
+
+          after do
+            MyActiveRecordModelTransition.updated_timestamp_column =
+              MyActiveRecordModelTransition::DEFAULT_UPDATED_TIMESTAMP_COLUMN
+          end
```
hmac: Think you should be able to do
```ruby
around do |example|
  MyActiveRecordModelTransition.updated_timestamp_column.tap do |original|
    MyActiveRecordModelTransition.updated_timestamp_column = nil
    example.run
    MyActiveRecordModelTransition.updated_timestamp_column = original
end
```

Unsure if that's any neater though...

lib/statesman/adapters/active_record.rb
timrogers: Nope!

spec/statesman/adapters/active_record_spec.rb
timrogers: Yes - great shout!

spec/statesman/adapters/active_record_spec.rb
timrogers: I think that's nicer as it clearly relates the before and after actions. Will do that now.

timrogers: I haven't managed to do this. It's working for the `"with no updated timestamp column set"` context (which doesn't use migrations) but not for the `"with a custom updated timestamp column set"` context. 

The issue I'm seeing is that Rails thinks the column has changed (even when you call `.reset_column_information`), but SQLite complaints that the column doesn't exist. I'm going to try a different approach of creating the `updated_on` column by default, and only using it in some tests.

timrogers:
@hmac Done!

timrogers:
@hmac Could you give this a final review?

PR #307, opened 2018-02-12
v3.3.1

timrogers:
- When unsetting the `most_recent` flag during a transition, don't assume that transitions have an `updated_at` attribute, but rather allow the "updated timestamp column" to be re-configured or disabled entirely (patch by [@timrogers](https://github.com/timrogers))


hmac:
This feels like a minor version change (it's kinda a new feature that we're adding). Thoughts?

timrogers:
Bit of a difficult one - the primary goal is to resolve a bug (https://github.com/gocardless/statesman/issues/305) but we add some new functionality of a sort in the process. What do you think?

hmac:
For me personally, if I see a minor bump then I expect nothing externally facing to change (unless a bug has been fixed). If we're having to update the documentation, that feels like it qualifies as a minor change?

timrogers:
Okay - let's describe it as v3.4.0 then.

timrogers:
Closing in favour of #308

PR #310, opened 2018-02-13
Support ActiveRecord transitions which don't include Statesman::Adapters::ActiveRecordTransition

### LINKED ISSUE #309: Introduction of `updated_timestamp_column` throws an exception when using the Postgres JSON column type for metadata
With v3.4.0, the addition of https://github.com/gocardless/statesman/pull/306 makes transitions dependent on `Statesman::Adapters::ActiveRecordTransition`, so [the advice given in the README when using a PostgreSQL JSON column type](https://github.com/gocardless/statesman#using-postgresql-json-column) to remove the include causes the following exception:

```ruby
NoMethodError:
       undefined method `updated_timestamp_column' for #<Class:0x00007fdd2ade07b0>
```

a workaround is obviously to now to explicitly set the `updated_at` in the transition class, which isn't ideal. 

```ruby
class_attribute :updated_timestamp_column
self.updated_timestamp_column = :updated_at
```

timrogers:
In #306, we made the "updated timestamp column" for a transition configurable through a `class_attribute` on `Statesman::Adapters::ActiveRecordTransition`, which is generally `include`d on transition classes.

However, not all transition classes include this module, and indeed we have [advised][1] in the README not to include it when using a PostgreSQL JSON column type for metadata, as pointed out by #309.

This leads to an exception like the following when trying to perform a transition:

```
NoMethodError: undefined method `updated_timestamp_column' for #<Class:0x00007fdd2ade07b0>
```

This fixes the issue by defaulting to using `:updated_at` as before if the transition class doesn't have an `.updated_timestamp_column` method defined.

As a follow-up action, it'd be nice to get all transitions including the module to have a more consistent and friendly base for future improvements to Statesman, as discussed in the [issue comments][2]. This, however, would be a breaking change and should be introduced separately after more though.

Fixes #309.

[1]: https://github.com/gocardless/statesman#using-postgresql-json-column
[2]: https://github.com/gocardless/statesman/issues/309#issuecomment-365336286

### DIFF
--- a/spec/statesman/adapters/active_record_spec.rb
+++ b/spec/statesman/adapters/active_record_spec.rb
@@ -163,6 +163,19 @@ describe Statesman::Adapters::ActiveRecord, active_record: true do
             to(change { previous_transition.reload.updated_at })
         end
 
+        context "for a transition class without an updated timestamp column attribute" do
+          let!(:adapter) do
+            described_class.new(MyActiveRecordModelTransitionWithoutInclude,
+                                model,
+                                observer)
+          end
+
+          it "defaults to touching the previous transition's updated_at timestamp" do
+            expect { Timecop.freeze(Time.now + 5.seconds) { create } }.
+              to(change { previous_transition.reload.updated_at })
+          end
+        end
+
         context "with a custom updated timestamp column set" do
           around do |example|
             MyActiveRecordModelTransition.updated_timestamp_column.tap do |original_value|
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Order.not_in_state(:checking_out) # => [#<Order id: "123">]
 
 By default Statesman stores transition history in memory only. It can be
 persisted by configuring Statesman to use a different adapter. For example,
-ActiveRecord within Rails:
+for ActiveRecord within Rails:
 
 `config/initializers/statesman.rb`:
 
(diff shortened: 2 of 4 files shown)


timrogers:
@mbaird Would you be able to give this fix a try?

hmac:
I wonder if it's worth mentioning in the docs around `updated_timestamp_column` that you can't use it (and we'll always default to `updated_at`) if you're using a json column?

timrogers:
Maybe - you could define it yourself though if you wanted?

mbaird:
@timrogers this works 👌🏼

timrogers:
@mbaird Thanks for letting me know! It works in a project we have too which is missing the `include` (not for any particular reason, mind you!).

timrogers:
I'm going to add a further commit with a `TODO` comment, and tweak the readme to make it clear that you should define `.updated_timestamp_column` if you want to customise the column and are not including the module.

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -163,6 +163,19 @@
             to(change { previous_transition.reload.updated_at })
         end
 
+        context "with a transition class without a updated timestamp column attribute" do
```
hmac: `an updated`

mbaird:
Thanks for the quick fix! 🙌🏼

PR #323, opened 2018-06-14
Add SemVer compatibility badge to README

greysteil:
Would you folks be up for showing your SemVer compatibility in a badge on your readme? We've just added the ability to do that using data from the test-runs of all the pull requests Dependabot creates - what do you reckon?

Here's what the badge will look like, and where it will link to:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=statesman&package-manager=bundler&version-scheme=semver)

Any feedback super appreciated!


greysteil:
@nickcampbell18 - any thoughts?

greysteil:
Updated the text of the badge to `semver stability`, which should make the use case clearer. Any thoughts?

nickcampbell18:
No strong feelings. @hmac, thoughts?

hmac:
Yeah, why not? I assume the number is rounded to the nearest 10%?

greysteil:
It’s exact - just happens that statesman is at a round number at the moment

greysteil:
@nickcampbell18 - updated!

hmac:
> Dependabot has updated statesman between SemVer compatible versions 39 times across 12 projects so far.
90% of those updates passed CI.

So 35.1 CI passes?

greysteil:
OK, so it's rounded to the nearest 1%. You know what I mean @hmac 😜

petehamilton:
@nickcampbell18 @hmac any idea what's going on with CI? They're all just pending but I can't find anything in CircleCI 🤔

This isn't going to break tests, so I'm just going to merge this in 🚢

PR #327, opened 2018-08-31
Make generators more rubocop friendly

greysteil:
- Statesman only support Ruby 2.2+, so we can use `%i(...)` syntax for symbol arrays
- Using single quotes and double quotes inconsistently was never necessary

Won't make the generators fit everyone's rubocop setup, but will mean they have a chance of fitting some people's.

Regen of #288 to try to fix CI.


nickcampbell18:
Hi @greysteil, looks like Statesman wasn't correctly configured to build forks & OSS 🤦‍♂️ . Would you mind force-pushing a new SHA to try and trigger a fresh build please?

greysteil:
Done. Doing better, but looks like that mongo image is missing.

danwakefield:
We've just added a new version so if you rebase it'll work now.

PR #342, opened 2019-04-09
Add better support for mysql in `transition_conflict_error?` method

greysteil:
Previously ~~the MySQL tests~~ all the transition conflict error tests were passing on Rails 6 because the index name included the bits we expecting.

In the wild, people don't always have nice names for their indexes, so relying on them including the names of the columns we're interested in isn't robust. This PR looks up the indexes properly.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -146,7 +146,24 @@ module Statesman
       end
 
       def parent_join_foreign_key
-        parent_model.class.to_s.demodulize.underscore + "_id"
+        association =
+          parent_model.class.
+            reflect_on_all_associations(:has_many).
+            find { |r| r.name.to_s == @association_name.to_s }
+
+        association_join_primary_key(association)
+      end
+
+      def association_join_primary_key(association)
+        if association.respond_to?(:join_primary_key)
+          association.join_primary_key
+        elsif association.method(:join_keys).arity.zero?
+          # Support for Rails 5.1
+          association.join_keys.key
+        else
+          # Support for Rails < 5.1
+          association.join_keys(transition_class).key
+        end
       end
 
       def with_updated_timestamp(params)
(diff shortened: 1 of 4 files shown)


greysteil:
@lawrencejones / @hmac can I get a review?

Should save you some pain when you upgrade to Rails 6. :octocat:

lib/statesman/adapters/active_record.rb
```diff
@@ -128,10 +128,27 @@ def serialized?(transition_class)
...
+          indexes(transition_class.table_name).
+          select do |index|
+            next unless index.unique
+
+            index.columns.sort == [parent_join_foreign_key, "sort_key"] ||
+              index.columns.sort == [parent_join_foreign_key, "most_recent"]
+          end
+      end
+
+      def parent_join_foreign_key
```
lawrencejones: Do we have any guarantee that people will name their columns like this? Seems like a more solid assumption than the one you've removed in this PR, but was wondering if we should do something smarter for this too?

greysteil: Good shout. Found a smarter way to do it.

greysteil:
@lawrencejones updated and working for all supported Rails versions. Can you take a second look?

lawrencejones:
I'm good with this but am away from a keyboard. I'll either wait for Harry to pick it up or handle it tomorrow morning when I'm back in work. Just want to try it out locally to double check everything before pressing merge!

greysteil:
Awesome, thanks @lawrencejones!

lawrencejones:
Minor bump sound sensible to you, @greysteil?

greysteil:
Yes please!

lawrencejones:
Should be released now: https://github.com/gocardless/statesman/releases/tag/v4.1.0. No doubt dependabot is busy opening PRs as we speak!

Issue #121, opened 2015-02-02
"in_state" scope takes long time to run


bfaguy:
we have a model called orderproduct that uses statesman.
whenever we perform a `Orderproduct.in_state()`, it is taking longer than expected to run. 
We have upwards of 700,000 to 800,000 records for this model. 
Is there anyway to speed up this scope?

thanks

versions
ruby '2.1.2'
gem 'rails', '4.1.2'
statesman (0.8.3)

isaacseymour:
Hi @bfaguy, this is an issue we ran into recently. We've internally implemented a solution which uses database triggers to maintain a `most_recent` flag on the most recent transition for each model. This means the query looks something like:
```sql
SELECT p.* 
FROM order_products p 
JOIN order_product_transitions t 
  ON t.order_product_id = p.id AND t.most_recent = 't' 
WHERE t.to_state = 'sent';
```
which is crazy fast when compared to the double-`LEFT OUTER JOIN` used by the current `in_state` implementation.

I'm working on a way of doing this within Statesman which is database-agnostic.

bfaguy:
Hi @isaacseymour Thank you for your quick response. 
Hope you can find a work around soon.

greysteil:
@bfaguy - https://github.com/gocardless/statesman/pull/132 is designed to address this (and is similar to the solution we're using internally).

bfaguy:
thank you @greysteil. We did find a workaround via our own code but your solution is greatly appreciated. @pdbradley

greysteil:
Closing now that #132 and #137 are included in a release.

PR #343, opened 2019-04-19
Use JOIN instead of LEFT OUTER JOIN for efficiency

josephfrazier:
We've found that these queries are causing a lot of latency, so I'm hoping they can be optimized by just JOINing instead of LEFT JOINing. See also https://github.com/gocardless/statesman/issues/121#issuecomment-72493310


josephfrazier:
Ah, I see that this would exclude the initial state. I wonder if there's another way to include that without having to do a LEFT JOIN.

danwakefield:
You could create the initial transition record at the point you create the parent model, using the `after_create_commit` or `after_create` hooks (depending on the guarantees you want) or inside `Actions::Foo:Create` if you use actions to centralise logic.

Another option would be to change `states_where` to completely special case the initial state and use `OUTER JOIN` only for that. I think this would be backwards compatible but I'd have to do some testing to be sure.

josephfrazier:
> Another option would be to change `states_where` to completely special case the initial state and use `OUTER JOIN` only for that. I think this would be backwards compatible but I'd have to do some testing to be sure.

Good call, I think I'll give this a shot.

josephfrazier:
> Another option would be to change `states_where` to completely special case the initial state and use `OUTER JOIN` only for that. I think this would be backwards compatible but I'd have to do some testing to be sure.

I finally got around to doing this. Could you take another look, @danwakefield?

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -30,6 +34,30 @@ def most_recent_transition_join
 
         private
 
+        def in_state_without_initial(*states)
+          states = states.flatten.map(&:to_s)
```
danwakefield:
```suggestion

```
Since this is a private method we can be sure that the callers are transforming the states into  strings above and remove the `map` here. The same applies to the `not_in_state` method.

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -30,6 +34,30 @@ def most_recent_transition_join
...
+
+          joins(most_recent_transition_join_without_initial).
+            where(states_where(most_recent_transition_alias, states), states).
+            where("#{most_recent_transition_alias}.to_state != ?", initial_state)
```
danwakefield:
```suggestion

```
You've done a check already that the states do not include `initial` to get to this point. Given that and the fact this uses `JOIN` I can't see a reason to add `!= inital_state`.
Maybe there is a reason I'm missing but if so It'd be great to comment this so the line isn't removed in the future.

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -30,6 +34,30 @@ def most_recent_transition_join
 
         private
 
+        def in_state_without_initial(*states)
+          states = states.flatten.map(&:to_s)
+
+          joins(most_recent_transition_join_without_initial).
+            where(states_where(most_recent_transition_alias, states), states).
```
danwakefield: It feels a bit weird to me that we are still using the `to_state IS NOT NULL` condition despite using `JOIN` which guarantees that not to be the case. Unless someone is using `to_state = NULL` on some records which honestly I don't think we should support.

lawrencejones:
Hey @josephfrazier, thanks for getting this up and running!

I've read through the changes and while it looks like they'd be functionally correct, I'm a bit unsure while changing from an `OUTER` join would make this more efficient.

Can you outline how you expect this to improve performance? The comment you linked to (https://github.com/gocardless/statesman/issues/121#issuecomment-72493310) refers to a time before Statesman managed a `most_recent` column, when the joins to find the most recent transition were quite expensive. Things are a lot better now given an index on `where most_recent = 't'`, so I'm not sure that comment applies in this situation.

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -30,6 +34,30 @@ def most_recent_transition_join
...
+          states = states.flatten.map(&:to_s)
+
+          joins(most_recent_transition_join_without_initial).
+            where("NOT (#{states_where(most_recent_transition_alias, states)})",
+                  states).
+            where("#{most_recent_transition_alias}.to_state != ?", initial_state)
+        end
+
+        def most_recent_transition_join_without_initial
+          "JOIN #{model_table} AS #{most_recent_transition_alias}
```
danwakefield: The meat of this method and `most_recent_transition_join` are the same apart from the `JOIN` condition at the beginning. Do you think it would make sense to have

```
def most_recent_transition_join
  "INNER JOIN " + join_statement
end

def most_recent_transition_join_without_initial
  "JOIN " + join_statement
end

private

def join_statement
  "#{model_table} AS #{most_recent_transition_alias}
     ON #{table_name}.id = #{most_recent_transition_alias}.#{model_foreign_key}	                  
     AND #{most_recent_transition_alias}.most_recent = #{db_true}"
end
```

to ensure they are always update in-sync.

josephfrazier:
Thanks for the thorough review, @danwakefield! I'll push some updates in a bit.

---

> Can you outline how you expect this to improve performance? The comment you linked to ([#121 (comment)](https://github.com/gocardless/statesman/issues/121#issuecomment-72493310)) refers to a time before Statesman managed a `most_recent` column, when the joins to find the most recent transition were quite expensive. Things are a lot better now given an index on `where most_recent = 't'`, so I'm not sure that comment applies in this situation.

Hmm @lawrencejones, the linked comment also uses a regular `JOIN` instead of the `LEFT OUTER JOIN`, so I'm hoping this change will still result in performance gains. Admittedly, it's just a hunch, but anecdotally we've seen improvements in our app that uses Statesman. I'll try to see if I can gather some more solid evidence.

lib/statesman/adapters/active_record_queries.rb
josephfrazier: Thanks, fixed in https://github.com/gocardless/statesman/pull/343/commits/d3d7ae4da4ea112c8a45705c91164cf6ade12ee8!

lib/statesman/adapters/active_record_queries.rb
josephfrazier: Hmm, interesting point. If we don't want to support `to_state = NULL`, we should make that (technically breaking) change separately, right?

lib/statesman/adapters/active_record_queries.rb
josephfrazier: Good catch! I was just kinda mindless copy/pasting, haha. Fixed in https://github.com/gocardless/statesman/pull/343/commits/4595bdcf7d13c2d4a34aa22e501f72a354376aaa

lib/statesman/adapters/active_record_queries.rb
josephfrazier: Good call, I fixed it in https://github.com/gocardless/statesman/pull/343/commits/5139e5d7ddf5a1139d85c6d12d234c5f9085c411, preserving the existing `LEFT OUTER JOIN` in the `most_recent_transition_join` method.

PR #347, opened 2019-05-16
Use OUTER JOIN only when necessary.

danwakefield:
The current behaviour of `LEFT OUTER` and `IS NOT NULL` when
querying states emulates a `JOIN`.
However databases can choose to optimize them differently.


danwakefield:
Out of date

PR #348, opened 2019-05-27
No easy way to find previous state

### LINKED ISSUE #263: No easy way to find previous state
Thanks for a great approach to a state machine. 

It seems like there is no easy method of finding the previous state for a state machine. I would have thought this would be something very common, so perhaps I am missing something? 

The `to_state` is captured in the DB for a transition, and I thought the `from_state` would either be saved in the DB, or deduced from transition information.

Is there a way of finding this easily? Or may you explain the reason this was excluded if it was done intentionally?

marclerodrigues:
Closes #263 

Before:
```
previous_state = project.transitions.where(most_recent: false).last.to_state || project.initial_state
```

After:
```
project.last_transition.from_state
```

### DIFF
--- a/lib/generators/statesman/templates/active_record_transition_model.rb.erb
+++ b/lib/generators/statesman/templates/active_record_transition_model.rb.erb
@@ -9,7 +9,7 @@ class <%= klass %> < <%= Statesman::Utils.rails_5_or_higher? ? 'ApplicationRecor
   # self.updated_timestamp_column = nil
 
 <%- unless Statesman::Utils.rails_4_or_higher? -%>
-  attr_accessible :to_state, :metadata, :sort_key
+  attr_accessible :from_state, :to_state, :metadata, :sort_key
 
 <%- end -%>
   belongs_to :<%= parent_name %><%= class_name_option %>, inverse_of: :<%= table_name %>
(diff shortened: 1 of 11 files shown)


dmagliola:
Could we have a `from_state` property on the Transition model, which does something like this, instead of adding a column?

```
def from_state
  transitions.where(id < self.id).last.to_state
end
```

Basically, i'm trying to avoid adding the redundant column. If the concern here is the performance hit of an extra DB call, rather than the non-obvious code one needs to write to get the "next-to-last" transition, then this obviously doesn't help.

marclerodrigues:
@dmagliola The point of this PR is getting the previous state without another DB call.

marclerodrigues:
@danwakefield I'm afraid this would be a breaking change since I'm making that a required value when transitioning.

What do you think about adding the `from_state` in the metadata? This way we wouldn't need a new column and would also save another DB call.

We could do that as a "workaround" and push it in a minor change and make the `from_state` column required in future major versions. Let me know your thoughts on this.

dmitry:
What about initial state?

dylan-hoefsloot:
@JoeSouthan I think this can be closed now that [Add from state method to transition history](https://github.com/gocardless/statesman/commit/40f9daed21980739b30d3aa4409e824b165775a4) has been merged.

PR #350, opened 2019-06-12
Inverse the order of writes on create transition

### LINKED ISSUE #206: most_recent column prevents HOT updates
https://github.com/postgres/postgres/blob/master/src/backend/access/heap/README.HOT

The statesman migration creates an index on the `most_recent` column. This prevents HOT updates on the transitions table.

The first solution that comes to my mind is to get rid of `most_recent` and instead have a `last_sort_key` on the parent model, then start every transition with an optimistic update of `last_sort_key`, which will lock the row for the rest of the transaction, which will prevent concurrent transitions, ensuring their unique ordering.

Both locking and retrieving all columns could even be done in a single statement: `UPDATE <model> SET last_sort_key = last_sort_key + 1 RETURNING *`.

arthurnn:
### Context

Dependabot is using statesman, and we are running the Dependabot service using MySQL at GitHub.
After some load testing the system, we found out some increasing number of deadlocks coming from states man. This fixed the problem for us.

This is a lock-free solution for #228. Probably also fixes #206

### Short problem description
When we were trying to call `ar_model.transition_to!`, we would get a Deadlock error if the system was under pressure. (lots of concurrent transitions happening at the same time) 

#### Some context
When doing a `transition_to!` statesman will do, mainly, [two operations](https://github.com/gocardless/statesman/blob/fdab116b9b0d9e1d77b5df151641761fbc7c15dd/lib/statesman/adapters/active_record.rb#L75-L82) in the database, inside the same transaction: 
1. Update the old transitions to mark them as non-most-recent
2. Insert the new transition with a most_recent set to TRUE. 

Also, the transitions table usually have two unique indexes, `(parent_id,sort_key)` and  `(parent_id,most_recent)`, the second index is there to guarantee we only have one most_recent transition for the parent entry. Thus we need the update before the insert.

#### Problem
MySQL default transaction isolation level is `REPEATABLE READ` and that's what we use at Dependabot/GitHub. [It states](https://dev.mysql.com/doc/refman/5.7/en/innodb-transaction-isolation-levels.html#isolevel_repeatable-read):
> This is the default isolation level for InnoDB. Consistent reads within the same transaction read the snapshot established by the first read.

What means is if in a loop, we do the same SELECT within a db transaction, even if other clients have committed data between the runs, it will, ALWAYS, see the first results it saw the first time.
That is also respected when doing writes (Update/Delete). so `UPDATE transitions SET most_recent=NULL where parent_id=1 AND most_recent=true;` will put an exclusive lock on the row for that unique index, as it cannot allow other transactions to change the data. However, if there is no row to update, MySQL will have to create a next-key lock, to also prevent others from changing this data, as the transaction's view of the snapshot didn't include any rows, as the UPDATE returned 0 rows changed.
The issue is, two independent transactions can put the next-key lock on the table (even if they reference different columns values on the index), and when it tries to do the next operation, the insert, it will lock, waiting for the next-key lock to be released. Therefore, if two updates happen first, and two inserts after, they will deadlock. (for better understanding see Simulation 1)
However, if the Update would have changed at least one row, there would be no locking, as for REPEATABLE READ that would be our current snapshot of the data. Thus no need for a gap locking. 
[See](https://dev.mysql.com/doc/refman/5.7/en/innodb-locking.html#innodb-gap-locks)
> If id is not indexed or has a nonunique index, the statement does lock the preceding gap.

This is also explained on this [2003 issue](https://bugs.mysql.com/bug.php?id=1866), by the people that wrote InnoDB, and on a few articles: [1](https://www.percona.com/blog/2013/12/12/one-more-innodb-gap-lock-to-avoid/) [2](https://www.percona.com/blog/2012/03/27/innodbs-gap-locks/)

### Solution

There are a few different ways to fix this. The easiest is just to (1)retry on deadlock, another one,(2) would be to change the transaction isolation level for that transaction to READ COMMITTED, which means SELECTS within the transaction are not consistent, and will always return committed data, thus, MySQL will not create the gap lock, as it is safe to update without the lock.
The problem of 1 is that we would still need to limit the number of retries, and eventually can still have the deadlocks, and number 2 would require us to always use ROW based replication as statement base replication wants REPEATABLE READ. [Also, not advised to change the transaction isolation level inside a library]
Instead, we can tackle the root problem: First, I thought in doing a select, and if transition data had to be updated do the update. However, that is race prone.
The actual fix is, re-order the operations, and always do the insert before, so we know the next Update will at least touch one row, and make our latest transition most_recent after. Thus, no next-key lock will happen and less contention will be created. 

To simulate this: 
### Simulation 1 (SQL)
A and B are different MySQL clients:
```
CREATE TABLE `transitions` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `to_state` varchar(40) NOT NULL,
  `sort_key` int(11) NOT NULL,
  `parent_id` bigint(20) NOT NULL,
  `most_recent` tinyint(1) DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_parent_sort` (`parent_id`,`sort_key`),
  UNIQUE KEY `index_parent_most_recent` (`parent_id`,`most_recent`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8mb4

A: START TRANSACTION;
B: START TRANSACTION;
A: UPDATE transitions SET most_recent=NULL where parent_id=1 AND most_recent=true;
B: UPDATE transitions SET most_recent=NULL where parent_id=42 AND most_recent=true;
A: INSERT INTO transitions(to_state, sort_key, parent_id, most_recent) values ('pending', 10, 1, TRUE);
B: INSERT INTO transitions(to_state, sort_key, parent_id, most_recent) values ('pending', 10, 42, TRUE);
```

cc @greysteil @hmarr

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -125,7 +125,7 @@ module Statesman
       def update_most_recents(most_recent_id)
         transitions = transitions_for_parent
         last_or_current = transitions.where(id: most_recent_id).or(
-          transitions.where(most_recent: true)
+          transitions.where(most_recent: true),
         )
 
         last_or_current.update_all(
@@ -246,7 +246,7 @@ module Statesman
         return nil if column.nil?
 
         [
-          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now,
+          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now
         ]
       end
     end
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -68,14 +68,16 @@ module Statesman
                                   sort_key: next_sort_key,
                                   metadata: metadata }
 
-        transition_attributes[:most_recent] = true
-
         transition = transitions_for_parent.build(transition_attributes)
 
         ::ActiveRecord::Base.transaction(requires_new: true) do
           @observer.execute(:before, from, to, transition)
-          unset_old_most_recent
+          # We save the transition first, and mark it as
+          # most_recent after to avoid letting MySQL put a
+          # next-key lock which could cause deadlocks.
           transition.save!
+          unset_old_most_recent
+          transition.update!(most_recent: true)
           @last_transition = transition
           @observer.execute(:after, from, to, transition)
           add_after_commit_callback(from, to, transition)
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -64,11 +64,9 @@ module Statesman
       private
 
       def create_transition(from, to, metadata)
-        transition_attributes = { to_state: to,
-                                  sort_key: next_sort_key,
-                                  metadata: metadata }
-
-        transition = transitions_for_parent.build(transition_attributes)
+        transition = transitions_for_parent.build(
+          default_transition_attributes(to, metadata),
+        )
 
         ::ActiveRecord::Base.transaction(requires_new: true) do
           @observer.execute(:before, from, to, transition)
@@ -84,6 +82,20 @@ module Statesman
         end
 
         transition
+      end
+
+      def default_transition_attributes(to, metadata)
+        transition_attributes = { to_state: to,
+                                  sort_key: next_sort_key,
+                                  metadata: metadata }
+
+        # see comment on `unset_old_most_recent` method
+        if transition_class.columns_hash["most_recent"].null == false
+          transition_attributes[:most_recent] = false
+        else
+          transition_attributes[:most_recent] = nil
+        end
+        transition_attributes
       end
 
       def add_after_commit_callback(from, to, transition)


greysteil:
@lawrencejones / @Sinjo - thought you might be interested in this one.

lawrencejones:
It's great what you find when you start running this stuff on different db engines! I hadn't considered MySQL and it's default of repeatable reads with this type of operation but it makes sense it won't be healthy for throughput. 

I think this probably makes sense but it does add a query to the transition operation which will have a measurable impact on people relying on low latency transitions. I'll put some time aside today to consider the full scope of that impact and give this a proper review.

danwakefield:
sqlite, postgres and mysql all support upsert, at least in recent versions.

Could we check for this and use it to update both of the columns
simultaneously?
We'd keep the current proposal as a fallback to maintain compatibility.

lawrencejones:
Using upsert is going to be really difficult through the AR layer- I think you'd need to insert the new transition, catch the conflict and write a fallback that both updates the conflicting transition and retries your initial insert. You'd also catch genuine race conflicts that should throw exceptions too, the sort you'd normally want to recover from with `retry_conflicts`.

Doing a general tidy-up of the statesman now to get a local development environment going, then I'm going to breakdown what SQL is actually being run here to be clear on how the behaviour will change. Will drop an update here when I've got something!

lawrencejones:
Hey @arthurnn, I had a think about this and have put together https://github.com/arthurnn/statesman/pull/1 that I'd love your feedback on. It's in the same spirit as your PR but removes the additional query which could be nice for those who care.

The description should explain everything but let me know if I can be clearer somewhere.

arthurnn:
@lawrencejones Pretty cool what you did on arthurnn#1, I like the approach of reusing the `UPDATE` to change the most_recent. Indeed is more complex. I don't mind the complexity though, however, not sure what the maintainers of the gem will think.
To be completely honest, I don't see the extra query(UPDATE) as a problem. It is equivalent to your UPDATE with the OR, the only thing, it adds an extra network call, and an exclusive lock for a bit longer.
Anyways, I am happy with either solution.

I am glad to merge your PR into my branch, so it will update this PR. just let me know.

@danwakefield  I thought about using an upsert, but we still need to update the old transitions to change the most_recent, thus I could not find a working solution using it.

danwakefield:
So because of `next_sort_key` the `@last_transition` contains its namesake until the new transition overwrites it after the save. This would give you access to both ids allowing you force an insert conflict.

But the `UPDATE` proposed by @lawrencejones makes sense and I think we are happy to add the complexity in order to support MySQL fully while being as performant as possible.

lawrencejones:
> the only thing, it adds an extra network call, and an exclusive lock for a bit longer

I suspect @greysteil has already mentioned this, but we use statesman at GoCardless for a load of our application code. In several places we transactionally transition many models and while the database queries are super-fast, the latency to the database dominates the query runtime. Consequently we'd much prefer to avoid adding extra hops if possible!

> I am glad to merge your PR into my branch, so it will update this PR. just let me know.

Let's do it!

Once we have that PR merged, we're happy to try this out in our staging environment to double check everything is sound. If things look good we'll :shipit:

arthurnn:
Interesting note on the db latency. 👍 

merged.

arthurnn:
will fix rubocop. one min

arthurnn:
`or` wont work on rails <5 =/

lawrencejones:
Just a note to say I'm not ignoring this, just busy today. Will get to fixing this next week.

lawrencejones:
A note to say I didn't get the time to look at things this week but intend to next. Hope it's not impacting anyone!

greysteil:
@lawrencejones gentle reminder on this one. @arthurnn is now on paternity leave so won't be responsive anymore 😢

lawrencejones:
Ahh that's really annoying, I missed getting to this before going on holiday last week. 

My general plan was to identify an alternative to using Arel OR, or maybe just waiting for Rails 6 to be released at which point we can deprecate support for older versions that don't support that functionality. 

I think the approach here (as in, the mechanics of how the SQL now work) is the correct one, just need to find a way to make tests pass. I'm going to be pressed for time this week (it's roadmap week internally) so would welcome any adjustments you'd want to make @greysteil?

lib/statesman/adapters/active_record.rb
```diff
@@ -96,21 +120,58 @@ def transitions_for_parent
         parent_model.send(@association_name)
       end
 
-      def unset_old_most_recent
-        most_recent = transitions_for_parent.where(most_recent: true)
+      # Sets the given transition most_recent = t while unsetting the most_recent of any
+      # previous transitions.
+      def update_most_recents(most_recent_id)
+        transitions = transitions_for_parent
+        last_or_current = transitions.where(id: most_recent_id).or(
```
greysteil: Equivalent SQL should be:

```
transitions.where("#{transition_class.table_name}.id = #{most_recent_id} OR #{transition_class.table_name}.most_recent = #{::ActiveRecord::Base.connection.quote(true)}")
```

The interpolation should be safe, too, since everything's from statesman internals.

We use a similar approach in `Statesman::Adapters::ActiveRecordQueries`.

Can PR it if you're 👍 on the approach.

lawrencejones: Absolutely- please do just add it to this PR!

Slightly concerned about `table_name` (if the table name is hyphenated, then I think this will break depending on what database you use?) but if we use it elsewhere then it's unlikely to be something people rely on.

greysteil: Think we use it elsewhere, but no reason not to quote the shit out of it 🤷‍♂

dmitry: Better to use AREL in this case, to be more safe and database agnostic.

lib/statesman/adapters/active_record.rb
```diff
@@ -96,21 +120,58 @@ def transitions_for_parent
...
+      end
+
+      # Check whether the `most_recent` column allows null values. If it doesn't, set old
+      # records to `false`, otherwise, set them to `NULL`.
+      #
+      # Some conditioning here is required to support databases that don't support partial
+      # indexes. By doing the conditioning on the column, rather than Rails' opinion of
+      # whether the database supports partial indexes, we're robust to DBs later adding
+      # support for partial indexes.
+      def not_most_recent_value
```
gabrielbaldao: What do you think about change this method to 
```ruby
false if transition_class.columns_hash["most_recent"].null == false
```

gabrielbaldao: What do you think about change this method to 
```ruby
false if transition_class.columns_hash["most_recent"].null == false
```

jurre:
Hey hey, we're still running into this issue, @arthurnn has switched teams but I'm happy to help wherever I can to get this over the line.

arthurnn:
I think this has being merged in master already right?

thom-oman:
@arthurnn yes, and we released it yesterday. this should've been closed, will close now.

PR #351, opened 2019-06-13
v6.0.0: Drop Ruby 2.2/2.3 and Rails 4.2 support

lawrencejones:
~~**We'll be unable to merge this until Rails 6 is released, as it otherwise drops support for a currently active Rails version (4.2). Check https://weblog.rubyonrails.org/releases/ for details.**~~

Rails 6 is released! let's get deprecating.

Ruby 2.2 reached EoL at 2019-03-31. As per our compatibility policy, we
should remove this version from our build matrix and cut a new major
Statesman release.

PR #352, opened 2019-07-02
Run tests against edgy Rails

isaacseymour:
- use 6.0.0.rc1, rather than 6.0.0.beta3
- also run tests against latest master of Rails (currently 6.1)
- define renamed method `trigger_transactional_callbacks?` required by 6.1 (see https://github.com/st0012/rails/commit/01f8abd7b2e46aa61eb60a96c456d67364948b24)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,6 +174,42 @@ jobs:
           - POSTGRES_USER=postgres
           - POSTGRES_DB=statesman_test
     steps: *steps
+  build-ruby262-rails-master-mongo:
+    docker:
+      - image: circleci/ruby:2.6.2-node
+        environment:
+          - RAILS_VERSION=master
+          - DATABASE_DEPENDENCY_PORT=27017
+      - image: circleci/mongo:3
+    steps: *steps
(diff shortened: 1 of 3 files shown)


lawrencejones:
It's worth calling out https://github.com/gocardless/statesman/pull/351 which would be relevant to this work. @isaacseymour let me know when you're finished breaking tests and I'll get to a review ;)

isaacseymour:
Ohh good point. Yep I'm done with breaking tests - I think the ones which are still broken are unrelated to 6.1, but break in 6.0 as well.

lawrencejones:
Hey @isaacseymour, I think these changes are still nice tidy-ups but will need rebasing against https://github.com/gocardless/statesman/pull/355

Do you want to do that then we can figure out which test failures are false positives/things we don't care about?

isaacseymour:
Rebased - seems like there's still that issue with mongo + sqlite3 + rails 6 🤔 v strange

isaacseymour:
Ok I think the issue is that [mongoid][1]

PR #354, opened 2019-07-04
Fix statesman index detection

hmarr:
The previous logic worked fine with parent tables that (approximately) began with the letters A-S, but with tables that (approximately) started with the letters T-Z, the comparison would fail, which ultimately resulted in the `RecordNotUnique` exception not being converted to a `TransitionConflictError`. Among other things, this prevents `retry_conflicts` from working.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -140,8 +140,8 @@ module Statesman
           select do |index|
             next unless index.unique
 
-            index.columns.sort == [parent_join_foreign_key, "sort_key"] ||
-              index.columns.sort == [parent_join_foreign_key, "most_recent"]
+            index.columns.sort == [parent_join_foreign_key, "sort_key"].sort ||
+              index.columns.sort == [parent_join_foreign_key, "most_recent"].sort
           end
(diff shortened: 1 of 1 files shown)


hmarr:
Looks like the mongo tests are failing due to https://github.com/rails/rails/issues/35153 (unrelated), and have been failing consistently for the past few builds.

hmarr:
Adding a test is a little tricky as the column name comes from the test model we use everywhere (`MyActiveRecordModel`). I did briefly consider renaming it to `ZyActiveRecordModel` but figured that'd confuse a lot of people 🤠 If you have any ideas on how we can easily test this without lots of stubbing and testing of private methods, I'm all ears! But it'd be great to get this fix in soon as we're seeing a _lot_ of exceptions because of it :-)

Good shout on the comment - added!

hmarr:
Thanks 🙌

Just rebased and pushed.

lawrencejones:
This should now be released in [v4.1.1](https://github.com/gocardless/statesman/releases/tag/v4.1.1)

hmarr:
Thanks Lawrence!

PR #355, opened 2019-07-04
Use Rails 6 rc1 in CI

hmarr:
Let's see if it fixes the build (https://github.com/rails/rails/issues/35153).


hmarr:
It doesn't fix it. What a pity! Still probably worth merging, though.

PR #359, opened 2019-07-25
Add text to clarify when a guard succeeds or fails

nacengineer:


README.md
```diff
@@ -234,7 +234,8 @@ end
 ```
 Define a guard. `to` and `from` parameters are optional, a nil parameter means
 guard all transitions. The passed block should evaluate to a boolean and must
-be idempotent as it could be called many times.
+be idempotent as it could be called many times. The guard will pass when the
```
martinstreicher: Does the guard pass if the result is “truthy” and fail if the result is “falsey”? I think this text would be clearer if it could compare apples and apples — right now, it seems to imply it must be a Boolean for truthy.

nacengineer: I'm fairly sure the guards require a boolean return. When true it triggers the guard. When false it doesn't. Which is my whole point in adding this text. Feel free to add whatever text you want. I just think there needs to be clarification as to what triggers a guard and what doesn't/

danwakefield:
Added in https://github.com/gocardless/statesman/pull/345, Thanks for the input that this needed to be clearer

PR #360, opened 2019-08-17
Add support for Rails 6 (and test it)

greysteil:
Rails 6 is here and Statesman doesn't yet support it. This fixes that.


greysteil:
@lawrencejones would you be up for getting this one out as a patch release? Appreciate that it will probably be followed by a major shortly after, but I'd like to unblock us getting onto Rails 6 as quickly as possible.

lawrencejones:
Yep, reckon this makes sense. The pending 5.0.0 release will be officially supporting Rails 6, along with applying our compatibility policy and removing support for Mongo, so if this is something you could do with a patch version for then I don't see why we can't get it out quickly now.

greysteil:
Legend, thanks!

lawrencejones:
Should be available in 4.1.2 now @greysteil https://github.com/gocardless/statesman/releases/tag/v4.1.2

PR #361, opened 2019-08-20
Add common delegations to README

danwakefield:
This makes it easier to copy and paste and give each machine a nice set of behaviour.


danwakefield:
Going to merge without the mongo tests because it's just a readme change and their is a PR in flight to drop mongo support completely.

PR #364, opened 2019-09-26
Fix typo in `check_missing_methods!`

thom-oman:
Small typo meant that this method wasn't actually checking if methods were implemented.

Fixes https://github.com/gocardless/statesman/issues/369

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -3,7 +3,7 @@ module Statesman
     module ActiveRecordQueries
       def self.check_missing_methods!(base)
         missing_methods = %i[transition_class initial_state].
-          reject { |_method| base.respond_to?(:method) }
+          reject { |m| base.respond_to?(m) }
         return if missing_methods.none?
 
         raise NotImplementedError,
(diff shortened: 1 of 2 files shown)


lib/statesman/adapters/active_record_queries.rb
```diff
@@ -3,7 +3,7 @@ module Adapters
     module ActiveRecordQueries
       def self.check_missing_methods!(base)
         missing_methods = %i[transition_class initial_state].
-          reject { |_method| base.respond_to?(:method) }
+          reject { |meth| base.respond_to?(meth) }
```
danwakefield:
```suggestion
          reject { |method| base.respond_to?(method) }
```

thom-oman: Yeah, I will do. I avoided `method` because it's an already defined method, I mean it won't have any negative affect but it has as little smell, what do you think?

danwakefield: Good point. `m` then? 
just think `meth` could be considered inappropriate

thom-oman: Fair enough, will change

PR #366, opened 2019-11-01
[feature] Add callbacks that trigger on transition failure

cedric-cordenier:
Hi there,

We're big fans of statesman at work, but we've found that we'd really appreciate the ability to be able to define callbacks that are triggered when a transition fails. In this PR, I've added a new `after_transition_failure` hook which extends the existing callback mechanism to add support for callbacks that are triggered if a transition fails.

I'd really appreciate your thoughts on this,

Thanks,

Cedric

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -48,6 +48,8 @@ module Statesman
         @callbacks ||= {
           before: [],
           after: [],
+          after_transition_failure: [],
+          after_guard_failure: [],
           after_commit: [],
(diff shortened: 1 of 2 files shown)


cedric-cordenier:
Hi @danwakefield -- sounds good. That's more obvious! I'll find some time to amend this PR tomorrow :)

cedric-cordenier:
@danwakefield I've updated my PR -- let me know if you have any further comments/questions

PR #367, opened 2019-11-06
Change transition failed error to explicitly include to / from states

ahjmorton:
Make the from / to states in the TransitionFailedError accessible to avoid parsing strings to determine which state transition failed.  

Allows the recipient to make more informed decisions when a state transition fails. For example if we're trying to transition to the same state due to a race condition we should take a different action than if we're transitioning to something that is invalid as a result of a programming error.

Maintain backwards compatibility with those who parsed the exception to get the transition states by maintaining the same error message.


ahjmorton:
@danwakefield for now I've reverted removed the versioning commit. Will re-add it on a tagged branch to release off of

PR #376, opened 2019-11-12
Change extend to include

Bramjetten:
This message just bit me when upgrading to Statesman 5.0.

The message says you need to extend Statesman::Adapters::ActiveRecordQueries, but you should use include instead.


danwakefield:
Good catch.

PR #378, opened 2019-11-13
Pass exception to after_failure callbacks

cedric-cordenier:
- The exceptions now contain attributes denoting the `from` and `to`
state which caused the exceptions. This is useful for the failure
callbacks as they don't currently know the state that was being
transitioned to.
- Pass the exceptions to the callbacks so they can make use of this
information.

### DIFF
--- a/lib/statesman/machine.rb
+++ b/lib/statesman/machine.rb
@@ -231,15 +231,20 @@ module Statesman
       @storage_adapter.create(initial_state, new_state, metadata)
 
       true
-    rescue TransitionFailedError
-      execute(:after_transition_failure, initial_state, new_state)
+    rescue TransitionFailedError => e
+      execute_on_failure(:after_transition_failure, initial_state, new_state, e)
       raise
-    rescue GuardFailedError
-      execute(:after_guard_failure, initial_state, new_state)
+    rescue GuardFailedError => e
+      execute_on_failure(:after_guard_failure, initial_state, new_state, e)
       raise
     end
 
-    def execute(phase, initial_state, new_state, transition = nil)
+    def execute_on_failure(phase, initial_state, new_state, exception)
+      callbacks = callbacks_for(phase, from: initial_state, to: new_state)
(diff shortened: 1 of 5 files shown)


cedric-cordenier:
Dan, I've updated the PR according to your suggestions :)

You'll also notice that I'm also passing through the from and to states when calling the guard callback. This is because the generated message would have been incorrect in cases where the guard was configured as a global guard (the `from` and `to` passed into the guard constructor would have been `nil` in this case).

lib/statesman/machine.rb
```diff
@@ -287,7 +292,12 @@ def validate_transition(options = { from: nil, to: nil, metadata: nil })
 
       # Call all guards, they raise exceptions if they fail
       guards_for(from: from, to: to).each do |guard|
-        guard.call(@object, last_transition, options[:metadata])
+        guard.call(
+          @object,
+          last_transition,
+          options[:metadata],
+          from: from, to: to,
```
danwakefield: I'm worried that this violates the upcoming change to kwargs (https://bugs.ruby-lang.org/issues/14183) 

While the change to error message is helpful since we don't currently provide details about `to` / `from` on the error object itself people may be parsing the message and changing that would be breaking.

cedric-cordenier: > While the change to the error message is helpful since we don't currently provide details about to / from on the error object itself people may be parsing the message and changing that would be breaking.

That's a very good point, Dan -- so if I understand your ask correctly, I should limit this PR to just points 1 and 2 above, and remove the changes which pass in transition information to the callback?

danwakefield: Yeah. I'll have a think about how to pass the information down and once I've figure that out I'll release a version bump that changes that message too.

cedric-cordenier: Sounds good 👍

cedric-cordenier:
@danwakefield Can I nudge you for a review?

PR #382, opened 2019-12-18
Clarify our versioning policy and CI configuration

nickcampbell18:
Following our [compatibility policy](https://github.com/gocardless/statesman/blob/master/docs/COMPATIBILITY.md), this PR:

* Deprecates Ruby 2.2 and 2.3 (both end-of-life)
* Updates the minor patch versions for 2.4 and 2.6
* Bumps the major version to 6.0 for the above.


nickcampbell18:
I'll do Ruby 2.7 support in a separate PR (because it's only `-rc1` at the moment) - should we wait to bump the major version after that?

danwakefield:
If we are only going to add it the the test harness I think it's ok to do a minor/patch bump since it can't affect the users of the library as long as the rest of the tests pass.

isaacseymour:
> I'll do Ruby 2.7 support in a separate PR (because it's only `-rc1` at the moment) - should we wait to bump the major version after that?

I think Ruby versions are normally released on Christmas day, so not long to wait to cut a release for 2.7?

lawrencejones:
Just realising this is exactly what I was trying to do in https://github.com/gocardless/statesman/pull/351

We can close that other PR and move with this one, but I do think this is a major version bump, as per our compatibility policy. We may not have broken compatibility with older versions now, but if we're removing them from our CI environment then there's no telling someone won't come along, write what they think is a minor feature that is incompatible with the old Ruby's, then release what is a breaking change under feature.

So I think this is a major bump, if that makes sense?

nickcampbell18:
Thanks all - let's get this out now. Ruby 2.7 and Bundler 2.1 incoming next!

PR #386, opened 2020-01-07
Ruby 2.7 support & fix deprecation warning

nickcampbell18:


lib/statesman/adapters/active_record.rb
```diff
@@ -197,8 +197,8 @@ def with_updated_timestamp(params)
     end
 
     class ActiveRecordAfterCommitWrap
-      def initialize
-        @callback = Proc.new
+      def initialize(&block)
+        @callback = Proc.new(&block)
```
isaacseymour: Could this not just be 
```suggestion
        @callback = block
```
?

nickcampbell18:
This seems to require Bundler 2.x. We can get there by dropping older Rails versions which are keeping bundler pinned down: https://github.com/gocardless/statesman/pull/388

danwakefield:
One small nitpick: We should also update the readme as it currently has `gem 'statesman', '~> 5.2.0'`

PR #388, opened 2020-01-08
Drop official support for Rails 4.2, 5.0 and 5.1

nickcampbell18:
As per our compatibility policy, this also bumps the major version (to `7.0`).


nickcampbell18:
(I'll remove the required CI statuses if we're happy with the change)

PR #353, opened 2019-07-03
Arthurnn/deadlocks

greysteil:
Adds a commit to https://github.com/gocardless/statesman/pull/350 (I'm not a maintainer, so couldn't do it from within that PR).

PR #399, opened 2020-04-29
Handle transitions differently for MySQL

thom-oman:
To avoid a gap lock with MySQL, first insert the new transition with most_recent=false and then flip the values for the latest two transitions in order.

Includes https://github.com/gocardless/statesman/pull/350 and https://github.com/gocardless/statesman/pull/353

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -16,6 +16,10 @@ module Statesman
         end
       end
 
+      def self.adapter_name
+        ::ActiveRecord::Base.connection.adapter_name.downcase
+      end
+
       def initialize(transition_class, parent_model, observer, options = {})
         serialized = serialized?(transition_class)
         column_type = transition_class.columns_hash["metadata"].sql_type
@@ -74,7 +78,7 @@ module Statesman
         ::ActiveRecord::Base.transaction(requires_new: true) do
           @observer.execute(:before, from, to, transition)
 
-          if db_mysql?
+          if mysql_gaplock_protection?
             # We save the transition first with most_recent falsy, then mark most_recent
             # true after to avoid letting MySQL acquire a next-key lock which can cause
             # deadlocks.
@@ -136,7 +140,7 @@ module Statesman
         # MySQL will validate index constraints across the intermediate result of an
         # update. This means we must order our update to deactivate the previous
         # most_recent before setting the new row to be true.
-        update.order(transition_table[:most_recent].desc) if db_mysql?
+        update.order(transition_table[:most_recent].desc) if mysql_gaplock_protection?
 
         ::ActiveRecord::Base.connection.update(update.to_sql)
       end
@@ -292,8 +296,8 @@ module Statesman
         ]
       end
 
-      def db_mysql?
-        ::ActiveRecord::Base.connection.adapter_name.downcase.starts_with?("mysql")
+      def mysql_gaplock_protection?
+        Statesman.mysql_gaplock_protection?
       end
 
       def db_true
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -127,7 +127,7 @@ module Statesman
       def update_most_recents(most_recent_id)
         transitions = transitions_for_parent
         last_or_current = transitions.where(id: most_recent_id).or(
-          transitions.where(most_recent: true)
+          transitions.where(most_recent: true),
         )
 
         last_or_current.update_all(
@@ -250,7 +250,7 @@ module Statesman
         return nil if column.nil?
 
         [
-          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now,
+          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now
         ]
       end
     end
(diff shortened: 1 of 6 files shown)


lib/statesman/adapters/active_record.rb
```diff
@@ -184,15 +284,45 @@ def with_updated_timestamp(params)
...
-                    else
-                      Time.now
-                    end
+        [
+          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now
+        ]
+      end
+
+      def db_mysql?
+        ::ActiveRecord::Base.connection.adapter_name.downcase.starts_with?("mysql")
```
jurre, line 296: Is there any way we could avoid special casing this for mysql?

If that's not possible could we make this a configuration option instead? Selfish reasoning: we use an internal adapter that is not called `mysql`. I can monkey patch this method internally, but if there is any way to avoid it that would be sweet.

thom-oman: I think that's a much better approach actually, makes this change opt-in which is great. Will try find time later today.

lawrencejones: On the not making it a special case, I did try implementing a platform agnostic strategy but the requirements of Postgres and MySQL were too difficult to get them to marry. Where MySQL could do X, Postgres would suffer, and vise versa.

Agree with what you're saying though, providing a configuration option for this would be best. We can even default it to true if you're MySQL, while giving you everything you need.

Thanks for the prompt!

jurre: Yeah that sounds great, and the explanation makes sense 👍 

Scope creep: at this point, should we consider creating separate `Statesman::Adapter`s (one for postgres, one for mysql) for the different AR adapters? That way we can avoid some of the branching that we see creeping in here.

lawrencejones: That might be a good idea, though if we're allowing this to be configurable, I guess you'd want something more like a mixin. `adapter.include(Statesman::Adapters::MySQLGapLock` or something.

That said, let's leave that until after we've got this in. I'm conscious of how long this has been hanging, very keen to deliver the value then clean-up!

thom-oman: Managed to get this passing this morning, took way longer than it should have as you can see with all the force pushes 🙈 lemme know what you think. I think config may need a small refactor next time we add one of these configuration options.

jurre: I think this looks great, thanks so much 🎉 ❤️

lib/statesman/config.rb
```diff
@@ -5,14 +5,19 @@
 
 module Statesman
   class Config
-    attr_reader :adapter_class
+    attr_reader :adapter_class, :gaplock_protection_enabled
 
     def initialize(block = nil)
+      @gaplock_protection_enabled = false
```
lawrencejones, line 11: We've mixed calling this gaplock_protection and mysql_gaplock_protection. Can we change things so we have one uniform reference, preferably `mysql_gaplock_protection` in all places?

lawrencejones: I think we should set this parameter to be `true` when the adapter looks like mysql. Otherwise all the mysql statesman users end up with a crucial performance optimisation not enabled, which sucks.

Wdyt?

spec/statesman/adapters/active_record_queries_spec.rb
```diff
@@ -20,7 +20,11 @@ def configure_new(klass, transition_class)
     prepare_other_model_table
     prepare_other_transitions_table
 
-    Statesman.configure { storage_adapter(Statesman::Adapters::ActiveRecord) }
+    Statesman.configure do
+      storage_adapter(Statesman::Adapters::ActiveRecord)
+      # These ENV vars are only set on the MySQL builds
+      mysql_gaplock_protection ENV.fetch("GAPLOCK_PROTECTION", false)
```
lawrencejones, line 26: We could avoid this if we default to true when we detect mysql

lib/statesman/config.rb
thom-oman: My thoughts on this were that we don't want people with postgres enabling this, so the API from the config perspective should make that clear. We do something similar in `storage_adapter`. No strong opinions though, did that make sense?

I agree we should enable this by default, I can't remember why I didn't.

lib/statesman/config.rb
```diff
@@ -5,14 +5,26 @@
...
-    attr_reader :adapter_class
+    attr_reader :adapter_class, :mysql_gaplock_protection
 
     def initialize(block = nil)
+      @mysql_gaplock_protection = db_mysql?
       instance_eval(&block) unless block.nil?
     end
 
     def storage_adapter(adapter_class)
       @adapter_class = adapter_class
```
lawrencejones, line 16: I _think_ we should probably set the mysql_gaplock_protection variable, if it's not nil, when we assign the adapter. I'm not sure it's fully safe for us to tap ActiveRecord::Base for this when we allow people to provide specific adapter classes, or have I misunderstood?

thom-oman: Not sure I fully understand what you're suggesting. Are you suggesting we defer `mysql?` logic to the adapter classes? Or just removing that logic from the initialiser. I agree that having AR specific stuff in the config isn't great.

thom-oman:
hey @jurre sorry for the further delays on this, still struggling to find the time so I thought I'd send an update haha. Seems the only remaining failures are with Rails master and postgres, some gem compatibility issues. Hoping it doens't take too long, but apologies again and thanks for your patience 🙏

greysteil:
Thanks @lawrencejones and @thom-oman!

jurre:
🎉 thanks so much @lawrencejones, @thom-oman, @arthurnn and @greysteil

Issue #406, opened 2020-08-11
ActiveRecord Adapter & Multiple Databases


ckorhonen:
I've recently updated to Rails 6.0 and have ran into an issue with this gem and the new support for multiple databases.

I have two databases in my application, and I want the transition class in my state machine to live in the secondary database.

Everything works, until I try and transition state:

``` ruby
foo.transition_to!(:bar)

   (0.4ms)  ROLLBACK
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "foo_transitions" does not exist
LINE 1: UPDATE "foo_transitions" SET "most_recent" = FALSE, "upda...
               ^
from /usr/local/bundle/ruby/2.6.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/postgresql_adapter.rb:675:in `exec_params'
Caused by PG::UndefinedTable: ERROR:  relation "foo_transitions" does not exist
LINE 1: UPDATE "foo_transitions" SET "most_recent" = FALSE, "upda...
```

It appears that Statesman is trying to write my transition class to the wrong database.

If I wrap the code, everything works correctly:

``` ruby
ActiveRecord::Base.connected_to(database: :secondary) do
     foo.transition_to!(:bar)
end
```

Models are setup as per the [Rails guide](https://guides.rubyonrails.org/active_record_multiple_databases.html), with everything residing in the secondary database inheriting from an abstract class.

Based on a cursory glance at the code it looks like `Statesman::Adaptors::ActiveRecord` is operating in the context of `ActiveRecord::Base.connection` rather than using whatever is utilized by the transition class.

ckorhonen:
Added a rough PR which appears to fix this - no regressions noted for any of my state machines, using either primary or secondary DBs.

robotex82:
We are having the same issues by upgrading from 3.4.1 to 10.0.0. Any chances to have this PR updated and merged?

stephenbinns:
Fixed in https://github.com/gocardless/statesman/pull/522

PR #407, opened 2020-08-11
Support Rails 6.0 Multi DB

ckorhonen:
This PR addresses the issue described in #406 - since Rails now supports multiple databases, we cannot assume that the transition class uses the connection as defined by `ActiveRecord::Base`.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -119,7 +119,7 @@ module Statesman
 
       def add_after_commit_callback(from, to, transition)
         @transition_class.connection.add_transaction_record(
-          ActiveRecordAfterCommitWrap.new do
+          ActiveRecordAfterCommitWrap.new(@transition_class.connection) do
             @observer.execute(:after_commit, from, to, transition)
           end,
         )
@@ -331,9 +331,9 @@ module Statesman
     end
 
     class ActiveRecordAfterCommitWrap
-      def initialize(&block)
+      def initialize(connection, &block)
         @callback = block
-        @connection = ::ActiveRecord::Base.connection
+        @connection = connection
       end
 
       def self.trigger_transactional_callbacks?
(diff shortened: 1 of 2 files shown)


thom-oman:
Hey @ckorhonen thanks for the PR. This looks pretty reasonable, happy to approve if this is ready. Think we may also need to change [this line](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record_queries.rb#L144) as well.

ckorhonen:
@thom-oman Good spot, I've updated that line.

The other thing I wasn't too sure about were the [class methods here](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L10-L21). Realistically I doubt the Rails 3 crutch is still needed, but it is possible that the different databases could use different adapters. 

Also not sure if I should be doing anything additional in terms of tests for this specific case? 

This has solved the issue in my own application - I'm using multiple state machines, with the transition classes residing in both the primary and secondary databases, no regressions encountered.

thom-oman:
Yeah IIRC we dropped support for Rails 3 earlier this year, don't have much context on that check but my gut tells me we should make the changes there too and we can remove Rails 3 support independently. The other method was added recently, is probably worth changing that too, but would only cause issues if the transitions table uses a different database (like MySQL instead of PG).

Regarding tests, yes I almost mentioned test, I struggled to think what the tests could look like such that they actually provide value. All scenarios I could think of would, in essence, testing Rails. Will give it me more thought and speak to some of my colleagues. Also will check our codebase for regressions.

ckorhonen:
Pushed some additional updates - as I was building out a state machine I noticed that `after_commit` wasn't working.

thom-oman:
Awesome. I think we should for sure be adding some tests for this then, so we can confirm that ActiveRecord is querying the correct database and that would've failed without the `after_commit` change.

thom-oman:
@ckorhonen any updates?

robotex82:
Hi, thank you all for the effort! Any chances to have this merged?

stephenbinns:
Superseded by https://github.com/gocardless/statesman/pull/522

Issue #408, opened 2020-08-13
Failure: calling update_most_recents without most_recent_id


ccvivien:
I'm moving my rails application from mysql to postgres, and everywhere I make a call to `transition_to`, it fails with the following error:
```
 TypeError:
       no implicit conversion of nil into String
```

the error occurs in `statesman-7.2.0/lib/statesman/adapters/active_record.rb`

`def build_most_recents_update_all_values(most_recent_id = nil)` on the line `Arel::Nodes::SqlLiteral.new(most_recent_value(most_recent_id))`

It seems to come from the version 7.2. I don't encounter such error with version 7.1

In the method `def create_transition(from, to, metadata)`, when not using mysql, it tries to call `update_most_recents` with no parameters, so that's why it fails building the arel sql literal.

thom-oman:
Hey @ccvivien sorry for the slow reply, not got round to looking at this properly today. Sounds like you may be in a weird state between PG and mysql. Will try dedicate sometime, but in the meantime could you tell me what `Statesman.mysql_gaplock_protection?` returns please?

ccvivien:
Hey @thom-oman !, no worries.

`Statesman.mysql_gaplock_protection?` returns `nil`

Normaly, I should have no trace left of mysql, I even removed `mysql2` gem from the Gemfile, and in `config/database.yml` there is no trace left aswell of the former mysql database.

PS: It should be one of my teammate who replies next as I'm absent for a week.

thom-oman:
Hey @ccvivien sorry for not getting back to you, something came up on Friday that prevented me from looking into this. 

Don't have all the context myself but I've taken a quick look at this this morning and my theory is that the error is due to the default value of the `most_recent` column _not_ being `false`. When we generate the migrations [here](https://github.com/gocardless/statesman/blob/master/lib/generators/statesman/templates/create_migration.rb.erb#L8) we set this default value to false based on the logic in [this method](https://github.com/gocardless/statesman/blob/master/lib/statesman/adapters/active_record.rb#L10). Which since you've only recently changed to postgres would've resulted in `most_recent` not having a default. 

I will try to confirm this my side later in the day, could you (or your teammate) see if changing this null value locally solves your problem? If this is the case, then we will have to add some documentation on transitioning from MySQL to Postgres to the README.

thom-oman:
Managed to reproduce this issue by removing the null value for the `most_recent` column. Will add a short paragraph in the readme for those migrating from mysql to postgres. 

[...]

PR #409, opened 2020-08-18
Use correct Arel for null

thom-oman:
In https://github.com/gocardless/statesman/issues/408 we saw that due to the changes in https://github.com/gocardless/statesman/pull/399 we weren't handling null values correctly when creating new transitions. Arel cannot cast `nil` to `null` and we see 
```
> Arel::Nodes::SqlLiteral.new(nil)
TypeError: no implicit conversion of nil into String
from (pry):1:in `initialize'
```

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -316,6 +316,10 @@ module Statesman
         ::ActiveRecord::Base.connection.quote(value)
       end
 
+      def db_null
+        Arel::Nodes::SqlLiteral.new("NULL")
+      end
+
       # Check whether the `most_recent` column allows null values. If it doesn't, set old
       # records to `false`, otherwise, set them to `NULL`.
       #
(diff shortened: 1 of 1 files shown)


lib/statesman/adapters/active_record.rb
```diff
@@ -316,17 +316,23 @@ def db_false
...
       # Check whether the `most_recent` column allows null values. If it doesn't, set old
       # records to `false`, otherwise, set them to `NULL`.
       #
       # Some conditioning here is required to support databases that don't support partial
       # indexes. By doing the conditioning on the column, rather than Rails' opinion of
       # whether the database supports partial indexes, we're robust to DBs later adding
       # support for partial indexes.
-      def not_most_recent_value
-        return db_false if transition_class.columns_hash["most_recent"].null == false
+      def not_most_recent_value(db_cast: true)
```
thom-oman, line 330: Looking for better ways to do this.

PR #413, opened 2020-08-27
Require ruby >=2.3 in gemspec

h0lyalg0rithm:
Since the library uses the safe navigation operator, it is only present in ruby 2.3 and above.


benk-gc:
Closing since this has now been resolved in later versions.

Issue #415, opened 2020-11-03
ActiveRecord #reload method should reset last_transition instance


dmitry:
While transaction happens a rollback, `@last_transition` stays with state value not committed to the the database, to check the real state of the object you can run `#reload` method of the record (so it may load fresh data from the database). 

Currently it's not resetting the `#last` transition, like:

```ruby
      def reload(*)
        @last_transition = nil
      end
```

Spec is probably the following:

```ruby
    it "resets last with #reload" do
      ActiveRecord::Base.transaction do
        adapter.create(:y, :z)
        raise ActiveRecord::Rollback
      end
      expect(adapter.last.to_state).to eq("y")
      model.reload
      expect(adapter.last.to_state).to eq("y")
    end
```

dmitry:
Solved via #417 PR.

PR #417, opened 2020-11-04
Add #reset method to state machine and adapter interfaces #415

dmitry:
It will allow to reset state machine `history` and `last` states with `#reload`.
Very useful in case of rollbacks.

Related to the issue  #415

### DIFF
--- a/spec/statesman/adapters/active_record_spec.rb
+++ b/spec/statesman/adapters/active_record_spec.rb
@@ -355,6 +355,18 @@ describe Statesman::Adapters::ActiveRecord, active_record: true do
     end
   end
 
+  it "resets last with #reload" do
+    model.save!
+    ActiveRecord::Base.transaction do
+      model.state_machine.transition_to!(:succeeded)
+      model.state_machine.current_state
+      raise ActiveRecord::Rollback
+    end
+    expect(model.state_machine.current_state).to eq('succeeded')
+    model.reload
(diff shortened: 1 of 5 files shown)


spec/statesman/adapters/active_record_spec.rb
```diff
@@ -355,6 +355,18 @@
     end
   end
 
+  it "resets last with #reload" do
+    model.save!
+    ActiveRecord::Base.transaction do
+      model.state_machine.transition_to!(:succeeded)
+      model.state_machine.current_state
```
thom-oman, line 362: Do we need this line?

thom-oman (commented):
This looks great, can we fix the rubocop issues and check the tests (again, I know this is annoying)

spec/statesman/adapters/active_record_spec.rb
dmitry: Yes, it caches the value to the `@last_transition` instance variable. I can add a comment and expectation if you think it's better to see it there.

dmitry:
Fixed rubocop offenses.

spec/statesman/adapters/active_record_spec.rb
thom-oman: Nono that's fine, I was being silly haha. I assumed it would be that but didn't see anything with a very quick cursory glance.

thom-oman (approved):

PR #422, opened 2020-12-22
Support relationships that doesn't use `id` as a Primary Key

emancu:
The convention of naming the Primary Key of a table as `id` is widespread 
on Rails applications and very useful. 
However, exceptions are also supported and widespread in Rails projects.

If we decide to model an _entity_ with a primary key named `external_id` instead of  just `id`,
the methods provided by Statesman::Adapters::ActiveRecordQueries stop working as expected.

Even worse, if the model has the column `id` and `external_id` it will use the `id` even when this column is not a primary key, which leads to a problem because the query result will be incorrect but it won't fail as an error.


thom-oman (approved):
Hey @emancu, thanks for this and happy new year! This looks great, will look to get this out today or tomorrow, time permitting!

Issue #424, opened 2021-01-11
#421 induced strange behaviour for first transition (not_most_recent_value & db_null methods)


Paul-Yves:
When doing a transition, I now (after updating to 8.0.0) have a "no implicit conversion of nil into String" error. I tracked it down to the first transition, inside module Statesman::Adapters::ActiveRecord when not_most_recent_value is null, `db_null` returns nil when it used to return "null" [before this PR](https://github.com/gocardless/statesman/pull/421).

I managed to patch this by reverting to the old implementation of `Arel::Nodes::SqlLiteral.new("NULL")` or by copying the methods above with a `::ActiveRecord::Base.connection.quote(type_cast(nil))`. 

The first solutions seemed the most logical since the comment of type_cast says `Type casting against a column is deprecated and will be removed in Rails 6.2.` and the original implementation of db_null did not type cast against a column.

However I am not sure to fully understand the situation so I would like to have your inputs before submitting a PR

thom-oman:
Hey @Paul-Yves, thanks for getting in touch. 
Yeah, this looks like this was me!

Paul-Yves:
Thank you for answering, I'll try my best : 
We are in a rails 6.0.3.4 app, ruby 2.7.2. The state machine is very straightforward, something like : 
```ruby
class OrderWorkflowStateMachine
  include Statesman::Machine

  state :new, initial: true
  state :state_1
  state :state_2
  state :state_3

  states.each do |state|
    transition from: state, to: states - [state]
  end
end
```
and I used a breakpoint when doing `transition_to(:state_1)` in a newly created instance (so in the new state without transition). I reproduced the issue both in the specs (rspec) and using the app in my local development mode.

Our database is PostgreSQL 10.15 and we use the pg gem.

Relevant part of our database scheme would be 
```sql
CREATE TABLE public.order_workflow_transitions (
    id bigint NOT NULL,
    to_state character varying NOT NULL,
    metadata text,
    sort_key integer NOT NULL,
    order_id bigint NOT NULL,
    most_recent boolean,
    created_at timestamp without time zone NOT NULL,
    updated_at timestamp without time zone NOT NULL
);
```

Paul-Yves:
Any news about this ? Couldn't we simply revert to the previous implementation of db_null, since it was not a type casting, the change seems irrelevant to the original issue (deprecation warning)

thom-oman:
[...]

PR #427, opened 2021-01-19
Fix for `no implicit conversion of nil into String`

thom-oman:
As part of https://github.com/gocardless/statesman/pull/421 we changed `db_null` in the AR adapter to type cast `nil`. This breaks for Postgres databases that have a nullable `most_recent` column. 

Fixes: https://github.com/gocardless/statesman/issues/424


thom-oman:
The pending specs are for the old rails master, but rails has switched to 'main' and they're still hanging around, will be fine in future PRs

PR #430, opened 2021-02-03
Fix logo in README

hmarr:
The logo is currently broken! This is a Very Important Fix. It was hosted on Cloud App, but they seem to have lost the file. It's now on GitHub's CDN, which should be a better home for it. Additionally, it should be retina-friendly now.


lawrencejones (approved):
This is a good catch, thanks for grabbing it! Dead logos are really sad.

PR #434, opened 2021-03-29
Use Ruby 2.7 for Rails main

isaacseymour:
If we're ok with this, I'll remove `build-ruby265-rails-master-*` from the required statuses, and add `build-ruby-270-rails-602-*` and `build-ruby-270-rails-main-*`.

I don't think this requires a version bump of Statesman because we don't guarantee any support for rails@main, and just test it to stay ahead of upcoming issues?


Sinjo (approved):
👍🏻  Makes sense to me. You're right about the reasoning here. We're using this as an early warning about breaking changes on Rails `main` for our own sake.

PR #438, opened 2021-06-07
Add Machine#last_transition_to

isaacseymour:


thom-oman (approved):
Can we add this to the README as well?

isaacseymour:
Good point on the readme - updated.

PR #446, opened 2021-12-24
Deprecate ActiveRecord::Base.default_timezone

floriandejonckheere:
Rails 7.0 [deprecates](https://github.com/rails/rails/commit/c6e4dbeebbfa4ea2b6d50a2ffb75f707b68aabf4) `ActiveRecord::Base.default_timezone` in favour of `ActiveRecord.default_timezone`. Rails 7.1 will remove this method entirely. To prevent deprecation warnings in 7.0, the gem should default to the new method.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -304,8 +304,15 @@ module Statesman
         return nil if column.nil?
 
         [
-          column, ::ActiveRecord::Base.default_timezone == :utc ? Time.now.utc : Time.now
+          column, default_timezone == :utc ? Time.now.utc : Time.now
         ]
+      end
+
+      def default_timezone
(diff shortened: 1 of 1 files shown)


shkm:
@thom-oman not quite sure who to ping so apologies if you're not the right one for this, but what's the state of statesman (heheh)? It'd be great if we could get this tiny change in.

thom-oman:
Hey @shkm! Yeah I'm still keeping an eye on statesman but still finding less time than I'd like. So it's definitely not been receiving the love it deserves over the past few months 😅  but leave that with me, I'm going to try get the ball moving on that today.

In terms of this PR, I will also release it today, so we can get rid of those annoying warnings! :) 

Apologies again for the slowness, think this PR in particular was missed over the christmas period.

thom-oman (approved):

PR #489, opened 2023-01-17
Feature/instrument statesman

theaxel:


lib/statesman/callback.rb
```diff
@@ -19,7 +19,16 @@ def initialize(options = { from: nil, to: nil, callback: nil })
     end
 
     def call(*args)
-      callback.call(*args)
+      ActiveSupport::Notifications.instrument "callback.statesman", {
+        subject: args.first.class.name,
+        subject_id: args.first.id,
```
stephenbinns, line 24: Can we be sure these will always respond to `id`?

PR #492, opened 2023-02-22
Add optional initial transition

dylan-hoefsloot:
This will allow the state machine to track the initial state for the machine, this is important for 4 reasons:
- It allows us to keep a record of how long the state machine spent in the initial state
- If the initial state configuration changes after the state machine transitions from the initial state we still have an accurate record of what the initial state was
- If the initial state configuration changes before the state machine transitions from the initial state we still have an accurate record of what the initial state is
- It allows us to utilise transition hooks for transitioning to the initial state


spec/statesman/adapters/active_record_queries_spec.rb
```diff
@@ -254,7 +254,7 @@ def self.initial_state; end
       end
 
       it "does not raise an error" do
-        expect { check_missing_methods! }.to_not raise_exception(NotImplementedError)
+        expect { check_missing_methods! }.to_not raise_exception
```
dylan-hoefsloot, line 257:
```
WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives, since literally any other error would cause the expectation to pass, including those raised by Ruby (e.g. NoMethodError, NameError and ArgumentError), meaning the code you are intending to test may not even get reached. Instead consider using `expect { }.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`.
```

Tabby (commented):
Looks reasonable to me. Needs updating to current revision though, so that the CI checks run

dylan-hoefsloot:
@Tabby Any chance we get this merged?

PR #493, opened 2023-02-22
Add from state method to transition history

dylan-hoefsloot:
This is a continuation of the work of [marclerodrigues](https://github.com/marclerodrigues) in [No easy way to find previous state](https://github.com/gocardless/statesman/pull/348).

One of the concerns was:
> I like the idea.
>
> Is it possible to test if this gracefully falls back when a table does not have the from_state column?
> It'd be great if we didn't make table migrations required.

This PR adds checks to see if the column exists before querying it, therefore users who have not added this column will be able to continue with existing behaviour. This means that this is not a breaking change.

Persisting this information means that accessing it is faster and easier and avoid problems such as mentioned in [this comment](https://github.com/gocardless/statesman/issues/263#issuecomment-858770174) and the issue containing it.


dylan-hoefsloot:
Hey @stephenbinns,

Any chance you could give this a look over

dylan-hoefsloot:
The rubocop warnings that are causing the tests to fail are false positives, there are a bunch of new issues today on the rubocop project highlighting the same issue with the new `1.6.0` version.

We can either wait for it to be fixed there, downgrade rubocop or ignore these warnings for now.

simmerz:
I'd love this to be merged in. Anything stopping it?

dylan-hoefsloot:
@simmerz When I left off, it just needed a review.

JoeSouthan (approved):
Ran this against one of our projects internally and looks good to me. Apologies for the delay and thanks for the contribution!

PR #497, opened 2023-03-07
Update rubocop and enable new cops

stephenbinns:


.rubocop.yml
```diff
@@ -5,6 +5,7 @@ inherit_gem:
 
 AllCops:
   TargetRubyVersion: 3.0
```
Tabby, line 7: This should be set to the lowest version of Ruby we support with this gem, i.e. 2.7
(See https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version)
```suggestion
  TargetRubyVersion: 2.7
```

Gemfile
```diff
@@ -11,5 +11,6 @@ elsif ENV['RAILS_VERSION']
 end
 group :development do
   # test/unit is no longer bundled with Ruby 2.2, but required by Rails
+  gem "pry"
```
Tabby, line 14: Probably ought to be `require: false` so it's not loaded when not needed
```suggestion
  gem "pry", require: false
```

Tabby (commented):
Looks good. Would it be worth adding in the suggested Rubocop extensions in this PR too (rubocop-rake and rubocop-rails), or would that be better in a separate one?

.rubocop.yml
stephenbinns: > Looks good. Would it be worth adding in the suggested Rubocop extensions in this PR too (rubocop-rake and rubocop-rails), or would that be better in a separate one?

Probably best to split out. not sure they add much value to this if im honest

Tabby (approved):

PR #498, opened 2023-03-07
Sorbet compatible queries

stephenbinns:
Including `ActiveRecordQueries` to your model can cause issues with type checkers such as Sorbet, this is because this technically is using a dynamic include, which is not supported by Sorbet.

e.g.

```ruby
class Order < ActiveRecord::Base
  has_many :order_transitions, autosave: false
  include Statesman::Adapters::ActiveRecordQueries[transition_class: OrderTransition, initial_state: :pending]

  def state_machine
    @state_machine ||= OrderStateMachine.new(self, transition_class: OrderTransition)
  end
end
```

Cannot be parsed by Sorbet - https://sorbet.org/docs/adopting#step-4-fix-constant-resolution-errors
> Sorbet cannot statically analyze a codebase that dynamically includes code. For example, [code like this](https://sorbet.run/#module%20A%3B%20end%0Amodule%20B%3B%20end%0A%20%20%0Adef%20x%0A%20%20rand.round%20%3D%3D%200%20%3F%20A%20%3A%20B%0Aend%0A%20%20%0Aclass%20Main%0A%20%20include%20x%0Aend) is impossible to statically analyze.
> Dynamic includes must be rewritten so that all includes are constant literals.

To avoid these issues you can instead include the `TypeSafeActiveRecordQueries` module and pass in configuration, for example:

```ruby
class Order < ActiveRecord::Base
  has_many :order_transitions, autosave: false
  include Statesman::Adapters::TypeSafeActiveRecordQueries
  configure_state_machine transition_class: OrderTransition,
                          initial_state: :pending
  def state_machine
    @state_machine ||= OrderStateMachine.new(self, transition_class: OrderTransition)
  end
end

### DIFF
--- a/lib/statesman/adapters/type_safe_active_record_queries.rb
+++ b/lib/statesman/adapters/type_safe_active_record_queries.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Statesman
+  module Adapters
+    module TypeSafeActiveRecordQueries
+      def configure_state_machine(args = {})
+        transition_class = args.fetch(:transition_class)
+        initial_state = args.fetch(:initial_state)
+
+        include(
+          ActiveRecordQueries::ClassMethods.new(
+            transition_class: transition_class,
+            initial_state: initial_state,
+            most_recent_transition_alias: try(:most_recent_transition_alias),
+            transition_name: try(:transition_name),
+          ),
+        )
+      end
+
+      def self.included(base); end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/statesman/adapters/type_safe_active_record_queries.rb [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+          ActiveRecordQueries::ClassMethods.new(
+            transition_class: transition_class,
+            initial_state: initial_state,
+            most_recent_transition_alias: try(:most_recent_transition_alias),
+            transition_name: try(:transition_name),
+          ),
+        )
+      end
+
+      def self.included(base); end
```
stephenbinns, line 20: ✂️

PR #499, opened 2023-03-10
Add the source location of the Guard callback to GuardFailedError

JoeSouthan:
Often it can be difficult to know _which_ guard failed, especially if
there are a lot of guards that may intersect.

```
Statesman::GuardFailedError: Guard on transition from: '' to '["bar"]' returned false
from /Users/joesouthan/foobar/app/state_machines/foobars.rb:98
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v10.1.0 10th March 2023
+
+### CHanged
+- Add the source location of the guard callback to `Statesman::GuardFailedError`
+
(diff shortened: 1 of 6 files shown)


Tabby (approved):
Nice, thanks :)
Looks good to me

suibh (approved):
Merge this immediately.

benk-gc (approved):
🌟

Tabby:
Possibly resolves #254

PR #503, opened 2023-03-14
Fixed issue with transitions using STI

pedro-pedrosa:
Prior to 7.2.0 the code to unset the `most_recent` column of old transitions when a new transition was created relied on `transitions_for_parent`, which would build the correct query in the case that the transition class was using single table inheritance.

However, after #399, the code no longer uses the parent association to create the query and the current custom query lacks this functionality that existed prior to 7.2.0.

This PR addresses this issue by adding an AND clause to the query to filter by transition type if the transition class contains an inheritance column.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -158,13 +158,24 @@ module Statesman
 
       def most_recent_transitions(most_recent_id = nil)
         if most_recent_id
-          transitions_of_parent.and(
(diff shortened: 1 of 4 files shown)


Tabby (approved):
Looks great, thanks for fixing this!

artyrcheek:
Would love this to get merged, currently being bitten by this. Thanks @pedro-pedrosa

PR #505, opened 2023-03-31
Make sure `nil` is accepted as cached value

zerc:
If there are no previous transitions `nil` will be returned and should
be cached. Otherwise, the caching doesn't work for the initial
transition.


Tabby:
Can you please add a test case that fails without this PR and passes with?

I can't see how this makes any meaningful difference in behaviour

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -308,9 +308,9 @@
       described_class.new(MyActiveRecordModelTransition, model, observer)
     end
 
-    before { adapter.create(:x, :y) }
-
```
zerc, line 312: In tests of the `#last` method we were always creating a transition. This is not how it happens in reality.

spec/statesman/adapters/active_record_spec.rb
```diff
@@ -376,6 +376,16 @@
...
+
+    context "without previous transitions" do
+      it "does query the database only once" do
+        expect(model.my_active_record_model_transitions).
+          to receive(:order).once.and_call_original
+
+        expect(adapter.last).to eq(nil)
+        expect(adapter.last).to eq(nil)
+      end
+    end
```
zerc, line 388: This test verifies that the caching of the `#last` method works even when there are no transitions been made yet.

lib/statesman/adapters/active_record.rb
```diff
@@ -65,18 +65,18 @@ def history(force_reload: false)
...
       end
 
-      # rubocop:disable Naming/MemoizedInstanceVariableName
       def last(force_reload: false)
         if force_reload
           @last_transition = history(force_reload: true).last
+        elsif instance_variable_defined?(:@last_transition)
+          @last_transition
         else
-          @last_transition ||= history.last
```
zerc, line 73: tl;dr; this caching approach doesn't work when `history.last` returns `nil`

---

When the object is just created and there are no sate transitions, `history.last` returns `nil` because:

```
transitions_for_parent.order(:sort_key) => []
transitions_for_parent.order(:sort_key).last => nil
```

and then it's stored in `@last_transition` variable. 

If we call `#last` again (which happens when there are multiple `guard_transition` against the initial state transition) this statement:

```
@last_transition ||= history.last
```

will call `history.last` again which would result in the database query (which will return zero rows).

zerc:
@Tabby I've added the test and some comments to clarify the problem.

Tabby (approved):
Thanks for the change and explanation. I see what's happening now, and this change looks good to me :)

PR #507, opened 2023-04-03
reload should clear the cache

zerc:
- Allow safely reset the adapter
- Reset the adapter on model reload
- Bump the version


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## v10.2.1 3rd April 2023
```
stephenbinns, line 1: Needs a rebase on this, I think we'd be proposing this to be `10.2.2`

lib/statesman/adapters/active_record.rb
```diff
@@ -76,7 +76,8 @@ def last(force_reload: false)
       end
 
       def reset
-        remove_instance_variable(:@last_transition)
+        remove_instance_variable(:@last_transition) \
+          if instance_variable_defined?(:@last_transition)
```
stephenbinns, line 80: Nit, would generally prefer the full `if` block here rather than escaping the line break - I think this is one of the cops in gc_rubocop too

```suggestion
        if instance_variable_defined?(:@last_transition)
          remove_instance_variable(:@last_transition)
        end
```

spec/statesman/machine_spec.rb
zerc, line 1: Not related to this PR it's just rubocop started to complain about this for some reason and I had to make the change.

lib/statesman/adapters/active_record_queries.rb
```diff
@@ -49,6 +49,14 @@ def included(base)
...
           define_in_state(base, query_builder)
           define_not_in_state(base, query_builder)
+
+          define_method(:reload) do |*a|
+            instance = super(*a)
+            if instance.respond_to?(:state_machine, true)
+              instance.state_machine.reset
+            end
+            instance
+          end
```
zerc, line 59: NOTE: this will only work if 

```
include Statesman::Adapters::ActiveRecordQueries
```

is applied to the target model.

stephenbinns (approved):
Thanks!

zerc:
PR is merged but for some reason it's not marked as such on Github 😭

PR #373, opened 2019-11-11
Revert "Revert "Cleaner ActiveRecordQueries mixin""

danwakefield:
Reverts gocardless/statesman#370

Readds these changes for a major release.

PR #511, opened 2023-05-26
fix: Ensuring inheritance

davegudge:
The call to `ensure_inheritance` can cause a stack overflow when used in combination with the `enumerize` gem and a standard inherited model (not STI - no type column defined on the table).

https://github.com/gocardless/statesman/pull/373#discussion_r465877454

There's no need to continue with the `ensure_inheritance` when using a standard inherited model; the statesman methods are already available on the subclass.

---

I originally implemented this fix for [v7](https://github.com/skillstream/statesman/commit/8ca396d7368170c68badd357679aed67cd8e9c0d), then cherry-picked for [v9](https://github.com/skillstream/statesman/commit/cb1286052ef9b15a890740180757d5d4f0555644), and now [v10](https://github.com/skillstream/statesman/commit/9c8b9a7ee71443a07856a12343db80c6216cd113).

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -39,7 +39,7 @@ module Statesman
         end
 
         def included(base)
-          ensure_inheritance(base)
+          ensure_inheritance(base) unless base.subclasses.none?
 
           query_builder = QueryBuilder.new(base, **@args)
 
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -39,7 +39,7 @@ module Statesman
         end
 
         def included(base)
-          ensure_inheritance(base) unless base.subclasses.none?
+          ensure_inheritance(base) if base.respond_to?(:subclasses) && base.subclasses.any?
 
           query_builder = QueryBuilder.new(base, **@args)
 


lib/statesman/adapters/active_record_queries.rb
```diff
@@ -39,7 +39,7 @@ def initialize(**args)
         end
 
         def included(base)
-          ensure_inheritance(base)
+          ensure_inheritance(base) unless base.subclasses.none?
```
stephenbinns, line 42: Generally think this reads better as an if as I don't have to double negate

```suggestion
          ensure_inheritance(base) if base.subclasses.any?
```

Also I think subclasses was only added in Ruby 3.1 so we may need to change this to support 3.x which is still currently supported

davegudge: > Generally think this reads better as an if as I don't have to double negate

Very good point 👍🏼 

> Also I think subclasses was only added in Ruby 3.1 

You're right: https://ruby-doc.org/core-3.1.0/Class.html#method-i-subclasses

`ActiveSupport` has a `DescendantsTracker` module which may be of help? It looks like it historically had a [`direct_descendants`](https://api.rubyonrails.org/v6.0.2.2/classes/ActiveSupport/DescendantsTracker.html#method-i-direct_descendants), but that method has since been removed from [`DescendantsTracker`](https://api.rubyonrails.org/classes/ActiveSupport/DescendantsTracker.html)

stephenbinns: Bit of a tricky one as Ruby 3.0 is EOL in ~3 months it does feel like something we could ignore relatively soon, can we just see if `base` responds to `subclasses` so it doesn't go 💥

davegudge: 👍🏼  So are you thinking: 

```suggestion
          ensure_inheritance(base) if base.respond_to?(:subclasses) && base.subclasses.any?
```

PR #515, opened 2023-08-02
Define state constants

### LINKED ISSUE #513: Feature proposal: Allow accessing states as state machine constant
Would be nice if `.state` method defined a constant using the state name. eg.:
```ruby
class OrderStateMachine
  include Statesman::Machine
  
  state :pending
  state :failed
end

OrderStateMachine::PENDING # => "pending"
OrderStateMachine::FAILED # => "failed"
```

This way we can detect typos and changed states early:
```
Order.new(state: :peding) # ok
Order.in_state(:peding) # ok

Order.new(state: OrderStateMachine::PEDING) # uninitialized constant OrderStateMachine::PEDING (NameError)
Order.in_state(OrderStateMachine::PEDING) # uninitialized constant OrderStateMachine::PEDING (NameError)
```

What do you think? Would this be useful for more people?

stephannv:
Closes #513


README.md
```diff
@@ -366,6 +366,16 @@ Machine.successors
...
 ```
 
+## Class constants
+Each machine's state will turn into a constant:
+```ruby
+Machine.state(:some_state, initial: true)
+Machine.state(:another_state)
+
+Machine::SOME_STATE #=> "some_state"
+Machine::ANOTHER_STATE # => "another_state"
```
stephenbinns, line 376: Generally think this is a good idea, would be nicer if the states were in their own module so this would become

```ruby
Machine::States::SOME_STATE
```

lib/statesman/machine.rb
```diff
@@ -163,6 +165,16 @@ def validate_from_and_to_state(from, to)
 
       private
 
+      def define_state_constant(state_name)
+        constant_name = state_name.upcase
+
+        if const_defined?(constant_name)
+          warn "Name conflict: '#{self.class.name}::#{constant_name}' is already defined"
```
stephenbinns, line 172: I think the potential for this happening is enough to raise an exception, that said it should be much less likely to occur if we include a module for the dynamic constants

README.md
stephannv: I'm not sure if it would be nicer because these constants are already inside the specialized class, the StateMachine, so it becomes  `PaymentStateMachine::PAID`, and I think it follows the current code organization, we have code like: 
- `PaymentStateMachine.initial_state`
- `PaymentStateMachine.states`
- `PaymentStateMachine.successors` 

And not:
- `PaymentStateMachine::States.initial_state`
- `PaymentStateMachine::States.all`
- `PaymentStateMachine::States.successors`

What do you think? If you are sure that it will better inside the States module, I can make the changes.

lib/statesman/machine.rb
stephannv: I agree.

README.md
stephenbinns: Thinking about it a little more I think you're right, its probably less unexpected to have the constant in the same module as the call to `state`. I also noticed this is pattern we've unofficially adopted in a few places

README.md
```diff
@@ -390,6 +390,16 @@ Machine.successors
 }
 ```
 
+## Class constants
+Each machine's state will turn into a constant:
```
stephenbinns, line 394:
```suggestion
Adding a state to a state machine will automatically create a constant for the value, for example:
```

README.md
```diff
@@ -390,6 +390,16 @@ Machine.successors
...
+```ruby
```
stephenbinns, line 395: Would be good to give an example of a simple state machine here, e.g.

```ruby
class OrderStateMachine
  include Statesman::Machine

  state :pending, initial: true
  state :checking_out
  state :cancelled

  # Constants created as a side effect of adding state
  transition from: PENDING, to: [CHECKING_OUT, CANCELLED]
end
```
Which shows how why the constants are more useful - i.e. error at app boot rather than runtime.

stephannv:
Thanks @stephenbinns for the code review, I think your suggestions improved a lot this pull request. If you have more suggestions, let me know.

JoeSouthan (approved):
Made some changes to do with the constants being redefined occasionally and tested on one of our projects.

Many thanks for the change, sorry about the long review time.

chrise86:
@JoeSouthan can you cut a new version of the gem with this change please?

JoeSouthan:
@chrise86 Thanks for the reminder, included in https://github.com/gocardless/statesman/releases/tag/v13.1.0

PR #521, opened 2023-11-22
Spike: Try multi-db support

ameykusurkar:
This is mostly replacing `ActiveRecord::Base` with the actual model, so that we connect to the right database.


benk-gc:
Changes implemented in #522

PR #522, opened 2023-11-23
Add multi-database support to Statesman.

benk-gc:
Statesman currently relies heavily on `ActiveRecord::Base`, either explicitly or implicitly, when querying database connections. This doesn't play well when we have models or transitions which live on a different database, since it forces us to open a query to the primary connection pool.

This is a series of changes to allow Statesman to use the context it has available for the model or transition class, and make use of the appropriate connection.

### DIFF
--- a/lib/statesman/adapters/active_record.rb
+++ b/lib/statesman/adapters/active_record.rb
@@ -7,17 +7,13 @@ module Statesman
     class ActiveRecord
       JSON_COLUMN_TYPES = %w[json jsonb].freeze
 
-      def self.database_supports_partial_indexes?
+      def self.database_supports_partial_indexes?(model)
         # Rails 3 doesn't implement `supports_partial_index?`
-        if ::ActiveRecord::Base.connection.respond_to?(:supports_partial_index?)
-          ::ActiveRecord::Base.connection.supports_partial_index?
+        if model.connection.respond_to?(:supports_partial_index?)
+          model.connection.supports_partial_index?
         else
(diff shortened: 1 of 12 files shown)


spec/statesman/adapters/active_record_spec.rb
```diff
@@ -343,12 +341,35 @@
         end
       end
     end
+
```
orlylevk, line 344: do we want to add in one of the current tests a check that it is connected to the primary?

orlylevk (approved):

lib/statesman/adapters/active_record.rb
```diff
@@ -7,19 +7,15 @@ module Adapters
...
-      def self.database_supports_partial_indexes?
+      def self.database_supports_partial_indexes?(model)
         # Rails 3 doesn't implement `supports_partial_index?`
-        if ::ActiveRecord::Base.connection.respond_to?(:supports_partial_index?)
-          ::ActiveRecord::Base.connection.supports_partial_index?
+        if model.connection.respond_to?(:supports_partial_index?)
+          model.connection.supports_partial_index?
         else
-          ::ActiveRecord::Base.connection.adapter_name == "PostgreSQL"
+          model.connection.adapter_name == "PostgreSQL"
```
orlylevk, line 15: it is the original code but not sure why we do not ignore case

ankithads (approved):

danielmamay (approved):

PR #525, opened 2023-11-30
Bump version to 12.0.0.

benk-gc:
- Added multi-database support [#522](https://github.com/gocardless/statesman/pull/522)
  - This now uses the correct ActiveRecord connection for the model or transition in a multi-database environment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v12.0.0 30th November 2023
+
+### Changed
+- Added multi-database support [#522](https://github.com/gocardless/statesman/pull/522)
+  - This now uses the correct ActiveRecord connection for the model or transition in a multi-database environment
+
 ## v11.0.0 3rd November 2023
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## v12.0.0 30th November 2023
+
+### Changed
+- Added multi-database support [#522](https://github.com/gocardless/statesman/pull/522)
```
ameykusurkar, line 4: I think this needs to include something along the lines of: "use the ActiveRecord connection of the transaction class etc", as that is the actual change. The multi-db support is more of a side effect of using the appropriate connection.

Do we need to update any documentation? Eg. any caveats users need to be aware of?

benk-gc: Good shout, updated. The usage and interface of statesman should be exactly the same from a model/transition point of view - the only thing that has changed is the connection used when creating and updating transitions.

benk-gc: Nope, only internal changes to behaviour detection methods.

ameykusurkar (approved):

PR #548, opened 2025-02-05
Add Ruby 3.4 support

JoeSouthan:
https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/

- Remove support for Rails < 7, add support for Rails 8
- Remove support for PostgreSQL < 13, add support for PostgreSQL 17
- Remove support for Ruby < 3.0, add support for Ruby 3.4


.github/workflows/tests.yml
```diff
@@ -24,18 +25,39 @@ jobs:
...
+            rails-version: "8.0"
+          - ruby-version: "3.1"
+            postgres-version: "15"
+            rails-version: "8.0"
+          - ruby-version: "3.1"
+            postgres-version: "16"
+            rails-version: "8.0"
+          - ruby-version: "3.1"
+            postgres-version: "17"
+            rails-version: "8.0"
```
Tabby, line 60: Why is `postgres-version` being added to these if we're adding exclusions for all postgres versions for each combination of ruby & rails that's being excluded?

Does it not meaningfully behave the same as this, but with far more checks needing to be run (and being significantly harder to read)
```suggestion
          - ruby-version: "3.1"
            rails-version: "main"
          - ruby-version: "3.1"
            rails-version: "8.0"
```

.github/workflows/tests.yml
```diff
@@ -69,18 +92,29 @@ jobs:
...
+          - ruby-version: "3.1"
+            mysql-version: "8.4"
+            rails-version: "main"
+
+          - ruby-version: "3.1"
+            mysql-version: "8.0"
+            rails-version: "8.0"
+          - ruby-version: "3.1"
+            mysql-version: "8.4"
+            rails-version: "8.0"
```
Tabby, line 116: Same as above, I don't see what the benefit is of having the mysql versions in the exclusions matrix if it's not being used to make new combinations
```suggestion
          - ruby-version: "3.1"
            rails-version: "main"
          - ruby-version: "3.1"
            rails-version: "8.0"
```

statesman.gemspec
```diff
@@ -20,21 +20,11 @@ Gem::Specification.new do |spec|
...
-  spec.add_development_dependency "rake", "~> 13.2.1"
-  spec.add_development_dependency "rspec", "~> 3.1"
-  spec.add_development_dependency "rspec-github", "~> 2.4.0"
-  spec.add_development_dependency "rspec-its", "~> 1.1"
-  spec.add_development_dependency "rspec-rails", "~> 6.0"
-  spec.add_development_dependency "sqlite3", "~> 1.7.0"
-  spec.add_development_dependency "timecop", "~> 0.9.1"
+  unless ENV["RAILS_VERSION"]
+    spec.add_dependency "rails", ">= 7.0"
+  end
```
Tabby, line 27: I believe it's possible to use Statesman without ActiveRecord, using the `Statesman::Adapters::Memory` adapter, so I don't think we should be adding a runtime requirement on ActiveRecord where there wasn't one before
```suggestion
```

Gemfile
```diff
@@ -10,7 +10,19 @@ elsif ENV['RAILS_VERSION']
   gem "rails", "~> #{ENV['RAILS_VERSION']}"
```
Tabby, line 10: Related to https://github.com/gocardless/statesman/pull/548/files#r1945092519, we should have a default version of Rails here for when `RAILS_VERSION` is not set, as we want to ensure there's always a Rails development dependency, but not necessarily a runtime dependency
```suggestion
  gem "rails", "~> #{ENV['RAILS_VERSION']}"
else
  gem "rails", ">= 7.0"
```

Issue #546, opened 2024-09-06
Transition migration generator doesn't run because model doesn't exist


simmerz:
When running `bin/rails g statesman:active_record_transition DirectDebitMandate DirectDebitMandateTransition` (yes, I'm making a model to hold a DD mandate using Statesman - very meta), the `DirectDebitMandate` model file already exists, but the `DirectDebitMandateTransition` one is due to be created.

However, the in the migration that creates the table (which happens *before* the model file is created), there is [this line](https://github.com/gocardless/statesman/blob/master/lib/generators/statesman/templates/create_migration.rb.erb#L8) which requires the model file to already exist.

My proposal would be to switch the order of the migration and model template lines in the [generator file](https://github.com/gocardless/statesman/blob/master/lib/generators/statesman/active_record_transition_generator.rb#L18) so that the file does exist when the migration is being created.

JoeSouthan:
Hi @simmerz 

We check the model because some of our models internally connect to different databases.

simmerz:
@JoeSouthan gotcha. So would a solution here be to check if the model exists, and if not to use the default connection?

PR #555, opened 2025-09-05
Fixup transition model generator for namespaced models

### LINKED ISSUE #460: Generator migration file name inconsistent with class name
# Description

Running the generator for a model under module results in inconsistent filename.

# Reproduction Instruction

`bundle exec rails g statesman:active_record_transition Game::Chess Game::ChessTransition`

# The error

The resulting filename would be `20220324024427_create_chess_transitions.rb`
And the resulting class name would be `class CreateGameChessTransitions < ActiveRecord::Migration[6.1]`

Running this migration results in error:
```
rails aborted!
NameError: uninitialized constant CreateChessTransitions
Did you mean?  CreateGameChessTransitions


Caused by:
NameError: uninitialized constant CreateChessTransitions
Did you mean?  CreateGameChessTransitions
```

# Expected outcome

The generated filename should be consistent with the classname.

`20220324024427_create_game_chess_transitions.rb`

JoeSouthan:
Fixes #460, #546

Namespaced models were not having their transition class generated correctly

Also fixes issue where the model is created after the migration.

Output:

```ruby
# 20250905103710_create_game_chess_transitions.rb
class CreateGameChessTransitions < ActiveRecord::Migration[8.0]
  def change
    create_table :game_chess_transitions do |t|
      t.string :from_state, null: false
      t.string :to_state, null: false
      t.jsonb :metadata, default: {}
      t.integer :sort_key, null: false
      t.integer :game_chess_id, null: false
      t.boolean :most_recent, null: false

      # If you decide not to include an updated timestamp column in your transition
      # table, you'll need to configure the `updated_timestamp_column` setting in your
      # migration class.
      t.timestamps null: false
    end

    # Foreign keys are optional, but highly recommended
    add_foreign_key :game_chess_transitions, :game_chesses

    add_index(:game_chess_transitions,
              %i[game_chess_id sort_key],
              unique: true,
              name: "index_game_chess_transitions_parent_sort")
    add_index(:game_chess_transitions,
              %i[game_chess_id most_recent],
              unique: true,
              where: "most_recent",
              name: "index_game_chess_transitions_parent_most_recent")
  end
end
```

### DIFF
--- a/lib/generators/statesman/generator_helpers.rb
+++ b/lib/generators/statesman/generator_helpers.rb
@@ -10,36 +10,44 @@ module Statesman
       "app/models/#{klass.underscore}.rb"
     end
 
-    def migration_class_name
-      klass.gsub("::", "").pluralize
-    end
-
-    def next_migration_number
-      Time.now.utc.strftime("%Y%m%d%H%M%S")
-    end
-
     def parent_name
-      parent.demodulize.underscore
+      parent.underscore.split("/").join("_")
     end
 
     def parent_table_name
-      parent.demodulize.underscore.pluralize
+      parent.underscore.split("/").join("_").tableize
     end
 
     def parent_id
       parent_name + "_id"
     end
 
-    def table_name
+    def association_name
       klass.demodulize.underscore.pluralize
+    end
+
+    def table_name
+      klass.underscore.split("/").join("_").tableize
+    end
+
+    def metadata_column_type
+      if ActiveRecord::Base.connection.supports_json?
+        postgres? ? :jsonb : :json
+      else
+        :text
+      end
     end
 
     def index_name(index_id)
       "index_#{table_name}_#{index_id}"
     end
 
+    def postgres?
+      configuration.adapter.try(:match, /postgres/)
+    end
+
     def mysql?
-      configuration.try(:[], "adapter").try(:match, /mysql/)
+      configuration.adapter.try(:match, /mysql/)
     end
 
     # [] is deprecated and will be removed in 6.2
--- a/lib/generators/statesman/generator_helpers.rb
+++ b/lib/generators/statesman/generator_helpers.rb
@@ -60,7 +60,7 @@ module Statesman
     end
 
     def database_supports_partial_indexes?
-      Statesman::Adapters::ActiveRecord.database_supports_partial_indexes?(klass.constantize)
+      Statesman::Adapters::ActiveRecord.database_supports_partial_indexes?(parent.constantize)
     end
 
     def metadata_default_value
--- a/lib/generators/statesman/generator_helpers.rb
+++ b/lib/generators/statesman/generator_helpers.rb
@@ -64,7 +64,13 @@ module Statesman
     end
 
     def metadata_default_value
-      Utils.rails_5_or_higher? ? "{}" : "{}".inspect
+      Utils.rails_5_or_higher? ? "{}" : "'{}'"
+    end
+
+    def metadata_column_config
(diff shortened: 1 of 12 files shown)


lib/generators/statesman/generator_helpers.rb
```diff
@@ -10,36 +10,44 @@ def model_file_name
...
+    def association_name
       klass.demodulize.underscore.pluralize
     end
 
+    def table_name
+      klass.underscore.split("/").join("_").tableize
+    end
+
+    def metadata_column_type
+      if ActiveRecord::Base.connection.supports_json?
```
Tabby, line 34: Should this not check the connection used by the model rather than `ActiveRecord::Base`, as some models might be using a connection to a different database?

lib/generators/statesman/templates/create_migration.rb.erb
```diff
@@ -1,9 +1,9 @@
-class Create<%= migration_class_name %> < ActiveRecord::Migration<%= "[#{ActiveRecord::Migration.current_version}]" if Statesman::Utils.rails_5_or_higher? %>
+class <%= migration_class_name %> < ActiveRecord::Migration<%= "[#{ActiveRecord::Migration.current_version}]" if Statesman::Utils.rails_5_or_higher? %>
   def change
     create_table :<%= table_name %> do |t|
       t.string :from_state, null: false
       t.string :to_state, null: false
-      t.text :metadata<%= ", default: #{metadata_default_value}" unless mysql? %>
+      t.<%= metadata_column_type %> :metadata<%= ", default: #{metadata_default_value}" unless mysql? %>
```
Tabby: nitpick/bikeshed: should we refactor the default value out into a method in `generator_helpers.rb` as well as `metadata_column_type`?

lib/generators/statesman/templates/update_migration.rb.erb
```diff
@@ -1,8 +1,8 @@
-class AddStatesmanTo<%= migration_class_name %> < ActiveRecord::Migration<%= "[#{ActiveRecord::Migration.current_version}]" if Statesman::Utils.rails_5_or_higher? %>
+class <%= migration_class_name %> < ActiveRecord::Migration<%= "[#{ActiveRecord::Migration.current_version}]" if Statesman::Utils.rails_5_or_higher? %>
   def change
     add_column :<%= table_name %>, :from_state, :string, null: false
     add_column :<%= table_name %>, :to_state, :string, null: false
-    add_column :<%= table_name %>, :metadata, :text<%= ", default: #{metadata_default_value}" unless mysql? %>
+    add_column :<%= table_name %>, :metadata, :<%= metadata_column_type %><%= ", default: #{metadata_default_value}" unless mysql? %>
```
Tabby: nitpick/bikeshed: should we refactor the default value out into a method in `generator_helpers.rb` as well as `metadata_column_type`?

lib/generators/statesman/generator_helpers.rb
Tabby: Actually nah this would be an edge case at best, and a maintenance nightmare if someone were to use different databases with different capabilities for different models so we can probably just say that's unsupported and leave this as-is

Tabby (approved):
Looks good to me. There's potential for a little more refactoring of the templates, but nothing essential

lib/generators/statesman/generator_helpers.rb
JoeSouthan: I think for generators it should be best effort, then it's up to the dev to fixup these details.

PR #560, opened 2026-01-05
Get test matrix up to date

jamiecobbett:
3.5-preview hasn't had a release since April 2025, and I suspect won't be released

https://www.ruby-lang.org/en/downloads/releases/

In any case, having this pre-release version is blocking upgrades to sqlite3: https://github.com/gocardless/statesman/actions/runs/20566094014/job/59064554664?pr=559

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,6 @@ jobs:
       matrix:
         ruby-version: ["3.2", "3.3", "3.4", "4.0"]
         rails-version:
-          - "7.1"
           - "7.2"
           - "8.0"
           - "8.1"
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -25,12 +25,15 @@ jobs:
...
-        ruby-version: ["3.2", "3.3", "3.4", "3.5"]
+        ruby-version: ["3.2", "3.3", "3.4", "4.0"]
         rails-version:
           - "7.1"
           - "7.2"
           - "8.0"
           - "main"
+        exclude:
+          - ruby-version: "3.2"
+            rails-version: "main"
```
Tabby, line 36: Presumably this is because of rails 8.1, so we should probably add "8.1" to the matrix explicitly as well

jamiecobbett: Rails 8.1 doesn't seem to require 3.3+, but it is worth adding 8.1 to the test matrix 👍

JoeSouthan (approved):

PR #564, opened 2026-07-16
Add configure_cached_current_state

JoeSouthan:
## Summary

- Adds `Statesman::Adapters::ConfigureCachedCurrentState`, a model mixin for maintaining the current state on a denormalised column on the parent model. Including the module does nothing on its own — the write is only wired up when `configure_cached_current_state(column:, touch_updated_at:)` is called explicitly.
- The write happens inside the `ActiveRecord` storage adapter itself, right after a transition is persisted as the new `most_recent` row. That single code path is shared by every `Machine` subclass and every adapter instance for a model, so it works even for models driven by several different machine classes chosen dynamically (e.g. per scheme), and fires reliably under `mysql_gaplock_protection`, where the `most_recent` flip happens via a raw SQL `UPDATE` that bypasses ActiveRecord callbacks entirely.
- It also runs before any of a machine's own `after_transition` callbacks for free, since it happens as part of persisting the transition, before Statesman invokes its own callbacks.
- `configure_state_machine` exposes `transition_class` (and `initial_state`) as class readers, so this mixin can read them instead of duplicating the same config.
- One consequence of the design: a transition row created directly on `transition_class`, bypassing the machine (e.g. in a backfill), does not update the cache — only real transitions performed through `transition_to!` do.

## Test plan

- [x] New specs cover: write on create, write on transition, custom column, touch_updated_at, missing transition_class raise, missing column raise, no-op until configured, ordering relative to a machine's own callbacks, and a model driven by more than one machine class.
- [x] Full existing suite still passes.

### DIFF
--- a/lib/statesman/adapters/active_record_queries.rb
+++ b/lib/statesman/adapters/active_record_queries.rb
@@ -33,6 +33,18 @@ module Statesman
         ClassMethods.new(**args)
       end
 
+      # Finds the has_many association on model that targets transition_class -
+      # matching by klass rather than by name, since callers (e.g.
+      # ConfigureCachedCurrentState) may not know what the association was named.
+      def self.transition_reflection_for(model, transition_class)
+        model.reflect_on_all_associations(:has_many).find do |reflection|
+          reflection.klass == transition_class
+        end || raise(
+          MissingTransitionAssociation,
+          "Could not find has_many association between #{model} and #{transition_class}.",
+        )
+      end
+
       class ClassMethods < Module
         def initialize(**args)
           @args = args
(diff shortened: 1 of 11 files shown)


lib/statesman/adapters/active_record_queries.rb
```diff
@@ -36,19 +36,31 @@ def self.[](**args)
...
+            transition_name: @args[:transition_name],
+          )
 
           base.define_singleton_method(:most_recent_transition_join) do
             query_builder.most_recent_transition_join
           end
 
           define_in_state(base, query_builder)
           define_not_in_state(base, query_builder)
+          define_cache_current_state(base) if @args[:cache_current_state]
```
orlylevk: Will this be done in the same DB transaction?

JoeSouthan: Yes — `after_transition` (the non-`after_commit` phase) runs inside the same DB transaction as the transition insert. `ActiveRecord::Adapters::ActiveRecord#create_transition` wraps `@observer.execute(:after, ...)` inside `transition_class.transaction(requires_new: true) { ... }` (lib/statesman/adapters/active_record.rb:87-114), and the cache-write callback is registered on the `:after` phase, so `update_columns` executes inside that block. If either the transition insert or the cache-column write fails, both roll back together.

(Note: this PR has since moved from `cache_current_state: true` on `configure_state_machine` to a separate `Statesman::Adapters::ConfigureCachedCurrentState` mixin / `configure_cached_current_state` call — same transactional guarantee applies, just a different call site.)

orlylevk (approved):

PR #566, opened 2026-07-28
Expose the transitioning object on GuardFailedError

shivanhi:
## Summary
- `Statesman::GuardFailedError` previously carried no reference to the object being transitioned, making it hard for callers to attach useful context (e.g. resource id/type) to error reports (Sentry, logs).
- `Guard#call` now passes the transitioning object through to `GuardFailedError.new`, and the error exposes it via a new `object` reader.
- The failure message itself is left unchanged (`"Guard on transition from: 'x' to 'y' returned false"`) — deliberately not interpolating the object into the message, since that would fragment Sentrygrouping (which keys off the message) and risks calling `#id` on an arbitrary caller-supplied object.
- README updated with an example of using `object` to attach resource context when notifying an error tracker.

## Test plan
- [x] `bundle exec rspec`
- [x] `bundle exec rubocop`

### DIFF
--- a/lib/statesman/guard.rb
+++ b/lib/statesman/guard.rb
@@ -5,8 +5,8 @@ require_relative "exceptions"
 
 module Statesman
   class Guard < Callback
-    def call(*args)
-      raise GuardFailedError.new(from, to, callback) unless super
+    def call(object = nil, last_transition = nil, metadata = nil)
+      raise GuardFailedError.new(from, to, callback, object) unless super
     end
   end
 end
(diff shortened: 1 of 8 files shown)


lib/statesman/guard.rb
```diff
@@ -6,7 +6,7 @@
 module Statesman
   class Guard < Callback
     def call(*args)
```
JoeSouthan: Can we not define the arguments here rather than splatting the args?

JoeSouthan (commented):
Seems good, can you add this to the changelog please?

lib/statesman/guard.rb
shivanhi: done

scyganska (approved):

JoeSouthan (approved):

